PR #1, opened 2022-02-04
Wrote injection function

wbenoit26:


alecgunny:
This looks really good Will! Couple things structural things that will help integrate better with the existing repo:

- I've added some pre-commit hooks to the repo that do things like code linting and formatting and import sorting to keep things aesthetically consistent. If you have Poetry installed on your development machine, you can run `poetry install` from the root directory, then `poetry run pre-commit install` which will make it so that these checks will run every time you try to commit your code, and will clean it up for you if possible before you can commit it (once installed, you can run `poetry run pre-commit run --all` to check your existing code). I have the repo configured to run these same checks before a PR can be accepted, so having them installed locally just makes life a little easier. I should have added these instructions to a README once I added them in, so apologies for that.
- You shouldn't need to configure `injectSignals.py` like a script (with `typeo` and the `if __name__ == "__main__":` syntax). The idea of it being in the `libs` section is that it's a library with injection-creating functions that can be called by other libraries or scripts to create arbitrary sets of signal-injected data.
- You'll need to add installation information about your library. In the context of Poetry, this means adding a `pyproject.toml` that includes dependencies described in the `[tool.poetry.dependencies]` section. You can see an example of this at [`libs/bbhnet/analysis/pyproject.toml`](./libs/bbhnet/analysis/pyproject.toml) and can find a more detailed description of what this file needs to contain [here](https://python-poetry.org/docs/pyproject/).
- There is at least some tension between the use of Poetry and the use of Conda in this repo, with Conda a necessary evil because some of gwpy's `.gwf` file I/O dependencies are only installable via Conda. We can chat in more detail about what my workaround for this looks like, but in the short term the basic idea is that I have a global Conda environment described in `environment.yaml` that installs all dependencies that require Conda, and then I use Poetry to install libraries into this environment (or clones of it). If that doesn't quite make sense right now, that's okay, but the upshot for you is that you can ignore specifying `gwpy` as a dependency in your `pyproject.toml` (and I think `h5py`  and `numpy` for that matter too), and just specify `scipy` and `bilby` (consult their PyPi pages for which minimum versions to specify).
- The way I have these libraries structured right now is that they all import from a dummy parent module called `bbhnet` (e.g. `from bbhnet.injection import ...`). You can see an example of how this looks in [`libs/analysis`](./libs/analysis), but the basic idea is that the top level of the library has a directory called `bbhnet`, which just contains a directory with the name of the library. This subdirectory is where all the library code ought to live (complete with a `__init__.py` file). So you might consider restructuring this as `libs/injection/bbhnet/injection/injection.py`, and renaming `main` in that file to `inject_signals`. Then in `libs/injection/bbhnet/injection/__init__.py`, including the line `from .injection import inject_signals` (and whatever other functions you want to expose to the public-facing API). Then a particular injection-generating application can list `libs/injection` as a dependency and do `from bbhnet.injection import inject_signals`.
- Adopting the previous structure involves adding one less-than-obvious line to your `pyproject.toml`. Under the `[tool.poetry]` section, add a `packages` field like you can find [here](https://github.com/alecgunny/bbhnet-monorepo/blob/988cd991079f9f46b75edc7e81fbf94de60a77ec/libs/analysis/pyproject.toml#L7-L9). This basically says to Poetry "Look for your expected directory structure under the `bbhnet` folder".
- More minor, but consider adopting [PEP8 naming practices](https://www.python.org/dev/peps/pep-0008/#naming-conventions). In particular, use `snake_case` with all lowercase letters (even for acronyms like SNR and GW) and underscores for all python file and function names, and reserve `CamelCase` (with a leading capital letter) for class definitions.
- At some point, it will be necessary to add unit tests under a `tests` directory, but since I've been lazy about writing these for the existing `analysis` and `io` libraries as well we can ignore it for now.

Feel free to continue to the discussion here, and
> use quote syntax
to talk about any of these points in particular. For anything you'd like to really get into the details of, Slack me or we can find some time to chat as well.

alecgunny:
I've also just noticed that you have an outdated version of the code in your fork. If you've added this repo as an `upstream` remote (if not, just run `git remote add upstream git@github.com:alecgunny/bbhnet-monorepo.git`), then you can do
```console
$ git checkout main
$ git pull upstream main
$ git checkout injection-function
$ git rebase main
```
Do this before you run any of the pre-commit hooks, as that will edit some of the existing files and cause conflicts since those files have also been edited in the upstream branch.

alecgunny:
Last thing: when you do pull from upstream, make sure to add your name and email to the author list in [`pyproject.toml`](pyproject.toml)!

wbenoit26:
Thanks for the comments, Alec! I'll take some time to address them today and let you know if I have questions.

alecgunny:
Yeah this was a feature added in Poetry 1.2, so it's most likely that you need to install the latest version of Poetry

```console
curl -sSL https://install.python-poetry.org | python3 - --preview
```

This is a bug on my end though because that should be required in the `[build-system]` section of `pyproject.toml` instead of `"poetry>=1.1"` so good catch, I'll fix this in `main`.

wbenoit26:
I believe I've addressed all the issues you raised, but let me know if there's anything more!

alecgunny:
This is looking way better! Some code I pushed this morning is making it fail the pre-commit checks, but if you rebase again on my most recent updates those should be fixed. That was bad code management on my part.

The one other procedural thing I'd note is that I think, as a rule, it's not desirable to set up the logging config within the context of a library. Applications should get decide how and where their application gets logged, and libraries are just responsible for sending messages to the logging machinery the application has set up. So I would remove those lines from the top of `injections.py`.

I'll take a closer look at the actual functionality this afternoon to see how this might integrate, but then we should be good to merge! Great work this is awesome

wbenoit26:
Sounds good! I've made those changes.

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,261 @@
...
+        duration=waveform_duration,
+        sampling_frequency=sample_rate,
+        frequency_domain_source_model=lal_binary_black_hole,
+        parameter_conversion=convert_to_lal_binary_black_hole_parameters,
+        waveform_arguments={
+            "waveform_approximant": "IMRPhenomPv2",
+            "reference_frequency": 50,
+            "minimum_frequency": 20,
+        },
+    )
```
alecgunny, line 175: So the assumption here is that the sample rate of the first frame file is the same as the sample rate for the rest of the frame files, correct?

Would it maybe make sense to have `sample_rate` be an argument to the function, then resample the strain timeseries to this rate upon reading? Or is it worth writing files "as-is", with the only difference being the injection of strain?

Alternatively, is this a time consuming step? Could we instantiate a new waveform generator for each file?

Otherwise, if this is going to be the most efficient way of doing things, it's probably worthwhile to insert a check for each file in case it doesn't match, to make sure we're not generating bad data.

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,261 @@
...
+
+    # sample GW parameters from prior distribution
+    priors = bilby.gw.prior.BBHPriorDict(prior_file)
+    sample_params = priors.sample(n_samples)
+
+    # Come back to this and see if can add frame_start
+    sample_params["geocent_time"] = signal_times
```
alecgunny, line 182: See my comment above about setting this on each iteration. I'm not super familiar with how bilby works, is having the absolute time important, or do you just need it relative to the start of your file?

If it is worthwhile to have the correct absolute time, I think we're ok to reset this with the appropriate `frame_start` at each iteration, right? I don't see where a frame's `"geocent_time"` param gets used again after its loop done.

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,261 @@
...
+    old_snr = np.sqrt(np.sum(np.square(snr_list)))
+    new_snr = np.random.uniform(snr_range[0], snr_range[1], len(snr_list[0]))
+
+    signals_list = [
+        signals * (new_snr / old_snr)[:, None] for signals in signals_list
+    ]
+    sample_params["luminosity_distance"] = (
+        sample_params["luminosity_distance"] * old_snr / new_snr
+    )
+    snr_list = [snr * new_snr / old_snr for snr in snr_list]
```
alecgunny, line 216: Can you explain what this block of code is doing? Do we need to do it before we write the injections to files? I can think of a lot of reasons that it would be nice to be able write injections as soon as we've created them, and not have to hold them all into memory at once.
- Allows us to do a lot more injections at once
- Allows us to not lose any data if anything goes wrong
- Saves us the extra loop below

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,261 @@
...
+        signal_strains, signals_list, frame_files
+    ):
+        for i in range(n_samples):
+            idx1 = int(
+                (signal_times[i] - waveform_duration / 2.0) * sample_rate
+            )
+            idx2 = idx1 + waveform_duration * sample_rate
+            strain[idx1:idx2] += signals[i]
+
+        strain.write(os.path.join(outdir, frame))
```
alecgunny, line 228: Presumably each `frame_file` will be a full path, right? Maybe something like `os.path.join(outdir, os.path.basename(frame))`?

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,261 @@
...
+        # write frame attributes
+        f.attrs.update(
+            {
+                "size": n_samples,
+                "frame_start": frame_start,
+                "frame_stop": frame_stop,
+                "sample_rate": sample_rate,
+                "psd_fftlength": fftlength,
+            }
+        )
```
alecgunny, line 257: Part of me thinks this metadata might make more sense to store with each individual injection, and save them as HDF5 files rather than GWFs.
- If we do ~1e4-1e5 injections, this file could become huge
- Generalizes better to downstream applications that might just need a random filename without having to worry about the intended structure of the directory that this file lives in
- It looks like this uses the frame start of the first file that got passed, but it's not obvious how other files could map that to their own frame start time, especially since we don't actually enforce that the first file in `frame_files` is chronologically the first.

Then again, I don't fully know how this info gets used so maybe there's good reasons to prefer this. I'm also sympathetic to the idea of leaving frame files "as-is", basically identical except that there's some injections in them.

The other thing that strikes me is that if the SNR re-weighting step above is absolutely necessary to do after all the individual SNRs have been calculated, we would have to loop through all the already written files and add the reweighted SNR value to the corresponding `ifo` group.

Thoughts here?

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,261 @@
...
+
+        # Total shift = shift to trigger time + geometric shift
+        dt = waveform_duration / 2.0
+        dt += ifo.time_delay_from_geocenter(ra, dec, geocent_time)
+        signal = np.roll(signal, int(np.round(dt * sample_rate)))
+
+        # Calculate SNR
+        if noise_psd is not None:
+            if get_snr:
+                snr[i] = get_snr(signal, noise_psd, sample_rate)
```
alecgunny, line 112: I believe the get_snr function can be array-ified and run at the end. I can take a crack at this if you'd like and submit it as a PR to your branch, which you can merge before I merge this into main.

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,261 @@
...
+
+        signals[i] = signal
+    if get_snr:
+        return signals, snr
+    return signals
+
+
+def inject_signals(
+    frame_files: [str],
```
alecgunny, line 121: Should be `Iterable[str]` (with `from collections.abc import Iterable` added to the top)

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,261 @@
...
+    channels: [str],
+    ifos: [str],
```
alecgunny, line 123: Will it be pretty common for every file to use the same channel and ifo? If so, it might make sense to make these `Union[str, Iterable[str]]`, and if the value passed is a string, just repeat it for the length of `frame_files`. Otherwise, check to ensure it has the same length as `frame_files`. (this would involve adding `from typing import Union` to the top)

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,261 @@
...
+    prior_file: str,
+    n_samples: int,
+    outdir: str,
+    fmin: float = 20,
+    waveform_duration: float = 8,
+    snr_range: [float] = [25, 50],
```
alecgunny, line 129: Either `Iterable[float]` or `Tuple[float, float]` to indicate that it should be of length 2

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,261 @@
...
+
+    sample_rate = waveform_generator.sampling_frequency
+    waveform_duration = waveform_generator.duration
+    waveform_size = int(sample_rate * waveform_duration)
+
+    signals = np.zeros((n_sample, waveform_size))
+    snr = np.zeros(n_sample)
+
+    ifo = bilby.gw.detector.get_empty_interferometer(ifo)
+    for i in range(n_sample):
```
alecgunny, line 77: It strikes me that this loop could probably be simplified by transposing this `dict` of `lists` to a `list` of `dicts` using the hacky but clever oneliner here https://stackoverflow.com/a/37489316

```python
sample_params = [dict(zip(sample_params, col)) for col in zip(*sample_params.values())]
```

This way the loop can just be
```python
for p in sample_params:
```
And you don't have to create a fresh dict each time. You could even do this at the top of the function and then just set `n_sample = len(sample_params)`.

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,261 @@
...
+        # Get parameter for one signal
+        p = dict()
+        for k, v in sample_params.items():
+            p[k] = v[i]
+        ra, dec, geocent_time, psi = (
+            p["ra"],
+            p["dec"],
+            p["geocent_time"],
+            p["psi"],
+        )
```
alecgunny, line 87: If you're just using these once for `ifo.antenna_response` and `ifo.time_delay_from_geocenter`, my inclination would be to skip this and just read from the dict during both those function calls (in keeping with the "three strikes you're out heuristic"), but that's primarily a matter a taste. If it makes those function calls look gross, then forget about it.

alecgunny: The only reason I bring it up is that this can be a bit confusing and make it appear like this is doing more than saving characters during the function calls below.

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,261 @@
...
+
+        # Generate signal in IFO
+        polarizations = waveform_generator.time_domain_strain(p)
+        signal = np.zeros(waveform_size)
+        b, a = sig.butter(
+            N=8,
+            Wn=waveform_generator.waveform_arguments["minimum_frequency"],
+            btype="highpass",
+            fs=waveform_generator.sampling_frequency,
+        )
```
alecgunny, line 97: Small thing, but this can be instantiated up front, right?

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,261 @@
...
+        for mode in polarizations.keys():
+            # Get ifo response
+            response = ifo.antenna_response(ra, dec, geocent_time, psi, mode)
+            polarizations[mode] = sig.filtfilt(b, a, polarizations[mode])
```
alecgunny, line 101: This can be done up front too, right? Then you could just have this loop look like:
```python
for mode, polarization in polarizations.items():
    response = ifo.antenna_response(ra, dec, geocent_time, psi, mode)
    signal += polarization * response
```

Or am I missing something and this actually gets set differently each time?

alecgunny (commented):
Last general question is, do you have any idea what the bottleneck step is here? Is it instantiating the waveform generator, sampling from the prior, generating the waveforms, computing the FFTs for the SNR? I was going to try to step through the code myself and see what I could find, but wondering if you already had at least some broad sense of this?

alecgunny:
Just went through the code in more detail. I think it has all of the high-level functionality and API that we want, but just have some implementation specific questions I'd be interested to get your thoughts on (and which will help me get a better sense for the nuts and bolts of how these injections get done). No rush, and obviously if you want to take some time to chat tomorrow or Thursday about all this to knock it all out in one fell swoop let me know.

libs/injection/bbhnet/injection/injection.py
alecgunny: Sorry this comment is meant to refer specifically to the line `polarizations[mode] = sig.filtfilt(b, a, polarizations[mode])`, which as I'm typing I'm realizing that these polarizations are being generated inside each loop so this can't be done up front. Still, I don't think resetting the key is necessary within the loop, and it might be cleaner as
```python
for mode, polarization in polarizations.items():
    response = ifo.antenna_response(ra, dec, geocent_time, psi, mode)
    signal += response * sig.filtfilt(b, a, polarization)

File: libs/injection/bbhnet/injection/injection.py [hunk 14]
```diff
@@ -0,0 +1,261 @@
...
+        np.random.choice(
+            np.arange(
+                waveform_duration,
+                frame_duration - waveform_duration,
+                waveform_duration,
+            ),
+            size=n_samples,
+            replace=False,
+        )
+    )
```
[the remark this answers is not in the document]
wbenoit26: Decided on desiring functionality for injecting sets of frames, so left as-is

libs/injection/bbhnet/injection/injection.py
wbenoit26: Added check to ensure all frames have the same sample rate

libs/injection/bbhnet/injection/injection.py
wbenoit26: Decided on desiring functionality for injecting sets of frames, so left as-is

libs/injection/bbhnet/injection/injection.py
wbenoit26: Decided on desiring functionality for injecting sets of frames, so left as-is

libs/injection/bbhnet/injection/injection.py
wbenoit26: Made this correction

libs/injection/bbhnet/injection/injection.py
wbenoit26: Opted against modification after conversation

libs/injection/bbhnet/injection/injection.py
wbenoit26: Fixed

libs/injection/bbhnet/injection/injection.py
wbenoit26: Fixed

libs/injection/bbhnet/injection/injection.py
wbenoit26: These changes make sense, and I've implemented them

libs/injection/bbhnet/injection/injection.py
wbenoit26: This is cleaner, and I've made this change

libs/injection/bbhnet/injection/injection.py
wbenoit26: Yes, and I've moved it outside the loop

libs/injection/bbhnet/injection/injection.py
wbenoit26: I'd appreciate if you could give this a try

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,285 @@
...
+    if len(span) != 1:
+        print("Error: The given strains have different durations")
+        exit()
+    frame_start, frame_stop = next(iter(span))
+    frame_duration = frame_stop - frame_start
+
+    sample_rate = set([int(strain.sample_rate.value) for strain in strains])
+    if len(sample_rate) != 1:
+        print("Error: The given strains have different sample rates")
+        exit()
```
alecgunny, line 162: Raise errors here, e.g. `raise ValueError("Frame files {} and {} have different durations".format(*frame_files))`

libs/injection/pyproject.toml
```diff
@@ -0,0 +1,18 @@
+[tool.poetry]
+name = "bbhnet.injection"
+version = "0.0.1"
+description = "Function for injecting BBH signals into input frames"
+authors = ["Will Benoit" <william.benoit@ligo.org,]
```
alecgunny, line 5: I think you need to move the close quote to the back and include a close caret > as well

libs/injection/bbhnet/injection/nonspin_BBH.prior
```diff
@@ -0,0 +1,15 @@
...
+ra = Uniform(name='ra', minimum=0, maximum=2 * np.pi, boundary='periodic')
+theta_jn = 0
+psi = 0
+phase = 0
+a_1 = 0
+a_2 = 0
+tilt_1 = 0
+tilt_2 = 0
+phi_12 = 0
+phi_jl = 0
```
alecgunny, line 15: I'm not sure this prior file makes sense to include in the library, since different applications will want to use their own sets of priors. This will definitely make sense to include in e.g. a project for making background and signal distributions on the same stretch of data

libs/injection/pyproject.toml
wbenoit26: Fixed

libs/injection/bbhnet/injection/nonspin_BBH.prior
wbenoit26: Good point, I've removed this

libs/injection/bbhnet/injection/injection.py
```diff
@@ -0,0 +1,285 @@
...
+            "minimum_frequency": 20,
+        },
+    )
+
+    # sample GW parameters from prior distribution
+    priors = bilby.gw.prior.BBHPriorDict(prior_file)
+    sample_params = priors.sample(n_samples)
+    sample_params["geocent_time"] = signal_times
+
+    signal_strains = []
```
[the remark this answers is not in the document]
wbenoit26: True. I've removed this

alecgunny (approved):
Looks great!

PR #18, opened 2022-03-09
glitch dataset generation

EthanMarx:


projects/generate-glitch-dataset/generate_glitches.py
```diff
@@ -63,18 +113,17 @@ def generate_glitch_dataset(
         # get path for this gps day 
         omicron_path = omicron_dir + f'{day}/'
         
-        # do we want pre or post vetoes? I presume we want to use some glitches found in vetoes (CAT3, not CAT1)
         trig_file = glob.glob(omicron_path + f'/*/PostProc/unclustered/triggers_unclustered_{ifo}.txt')[0]
```
alecgunny, line 116: consider more general path behavior here (i.e. either `os.path.join` or `Path(path_name) / "child_dir" / "file"`). Also consider documenting the expected structure of these directories, and maybe add a quick comment explaining what the structure of this file is so it makes the ensuing code using it more clear

EthanMarx:
@alecgunny So I altered the data discovery procedure from using fetch_open_data to gathering frames locally using gwdatafind on the LDG clusters for performance purposes. 

The test I wrote have passed locally, but will fail on the github servers since they don't have access to the frames. 

Any suggestions :)

projects/generate_glitches/generate_glitches.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+@typeo
+def main(
+    snr_thresh: float,
+    start: float,
+    stop: float,
+    window: float,
+    omicron_dir: str,
+    out_dir: str,
```
alecgunny, line 197: Minor point but you might consider using `pathlib.Path` annotations for these. `typeo` will pick that up and map them to `Path`s before passing them in, and then you can make use of all the nice `pathlib` syntax like using the `/` operator (not string) for separating path components and even the `Path.glob` method.

projects/generate_glitches/generate_glitches.py
```diff
@@ -0,0 +1,342 @@
...
+            (see /home/ethan.marx/bbhnet/generate-glitch-dataset/omicron/)
+    - channel: channel name used to read data
+    - frame_type: frame type for data discovery w/ gwdatafind
+    - H1_veto_file: path to file containing vetoes for H1
+    - L1_veto_file: path to file containing vetoes for L1
+    """
+
+    # create logging file in model_dir
+    logging.basicConfig(
+        filename=f"{out_dir}/log.log",
```
alecgunny, line 226: Here and in a few other places we make some assumptions about path separators. I don't ever really anticipate using this script on Windows, but at the very least it's inconsistent with also using `os.path.join`. This is primarily an aesthetic issue but it might make sense to opt for either one or the other (using `"/"` explicitly everywhere or providing every path component as an argument to `os.path.join`), or opting for using `pathlib` as outlined above.

projects/generate_glitches/generate_glitches.py
```diff
@@ -0,0 +1,342 @@
...
+
+    # TODO: I *Think* there is a pyomicron package
+    # that can create omicron dags. Might be useful
+    # to generalize this script to use pyomicron to produce
+    # omicron triggers for arbitrary stretches of data,
+    # channels, parameters etc..
+
+    gps_day_start = str(start)[:5]
+    gps_day_end = str(stop)[:5]
+    all_gps_days = np.arange(int(gps_day_start), int(gps_day_end) + 1, 1)
```
alecgunny, line 272: Minor point but would the same effect be achieved by ` // 10000` without having to go to strings and then back?

projects/generate_glitches/generate_glitches.py
```diff
@@ -0,0 +1,342 @@
...
+
+    H1_glitches = []
+    L1_glitches = []
+
+    H1_snrs = []
+    L1_snrs = []
+
+    # loop over gps days
+    for i, day in enumerate(all_gps_days):
```
alecgunny, line 281: I don't think you end up using `i` anywhere do you?

projects/generate_glitches/generate_glitches.py
```diff
@@ -0,0 +1,342 @@
...
+def veto(times: list, segmentlist: SegmentList):
+
+    """
+    Remove events from a list of times based on a segmentlist
+    A time ``t`` will be vetoed if ``start <= t <= end`` for any veto
+    segment in the list.
+
+    Arguments:
+    - times: the times of event triggers to veto
+    - segmentliss: the list of veto segments to use
```
alecgunny, line 38: segmentlist typo

projects/generate_glitches/generate_glitches.py
```diff
@@ -0,0 +1,342 @@
...
+
+    Returns:
+    - keep_bools: list of booleans; True for the triggers to keep
+    """
+
```
alecgunny, line 43: There's actually a pretty fun array-like way of doing this that I just tested if you're interested:
```
    mask = True
    for t0, tf in segmentlist:
        mask &= (t0 >= times) | (times >= tf)
    return mask
```

projects/generate_glitches/generate_glitches.py
```diff
@@ -0,0 +1,342 @@
...
+    # if passed, apply vetos
+    if vetoes is not None:
+        keep_bools = veto(triggers[:, time_col], vetoes)
+        triggers = triggers[keep_bools]
+
+    # restrict triggers to within gps start and stop times
+    times = triggers[:, time_col]
+    time_args = np.logical_and(times > start, times < stop)
+    triggers = triggers[time_args]
```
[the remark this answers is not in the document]
alecgunny: If you're interested in using the numpy loop above, you could simplify this whole section to just be:

```python
times = triggers[:, time_col]
mask = (times > start) & (times < stop)
if vetoes is not None:
    for t0, tf in vetoes:
        mask &= (t0 >= times) | (times >= tf)
triggers = triggers[mask]
```

projects/generate_glitches/generate_glitches.py
```diff
@@ -0,0 +1,342 @@
...
+
+    # apply snr thresh
+    day_snrs = triggers[:, snr_col]  # second column is snrs
+    snr_thresh_args = np.where(day_snrs > snr_thresh)
```
alecgunny, line 138: using implementation above, could include `mask &= triggers[:, snr_col] > snr_thresh`

projects/generate_glitches/generate_glitches.py
```diff
@@ -0,0 +1,342 @@
...
+        try:
+            glitch_ts = ts.crop(time - window, time + window)
+
+            glitch_ts = glitch_ts.resample(sample_rate)
+
+            snrs.append(trigger[snr_col])
+            glitches.append(glitch_ts)
+
+        except ValueError:
+            logging.info(f"Data not available for trigger at time: {time}")
```
alecgunny, line 182: maybe consider `logging.warning`, unless this is a pretty standard or expected occurrence

alecgunny (approved):
All my comments are entirely aesthetic: I was able to build the code and run the tests in a fresh environment, so I'd say this is good to go. The only one I would maybe push for would be removing or replacing the print statement, but otherwise everything else is up to you whether you want to merge.

projects/generate_glitches/generate_glitches.py
EthanMarx: @alecgunny Wow much simpler. I could be wrong but I think for vetoes files with many more segments this may take longer since each segment is checked every time for every trigger. Will keep this in mind though.

PR #20, opened 2022-03-09
Signal dataset generation

wbenoit26:


wbenoit26:
Let me know your thoughts on my testing - I've never written unit tests before. Happy to make them more exhaustive, or reformat, or whatever.

alecgunny:
This looks really good, couple points of feedback:

- Can any of the functionality in `simulate.py` use any of the functions you wrote for `libs/injections`? If so, consider adding that as a local dependency `"bbhnet.injection" = {path="../../libs/injection", develop = true}` in you `pyproject.toml`
- If all of your script's code can be fit into a single file, then consider just having a single file called `generate_waveforms.py` in the same directory as your `pyproject.toml` (no need for the nested structure)
- Consider adding your script as a command line executable in your poetry environment by adding `[tool.poetry.scripts]` section to your `pyproject.toml` and giving it a line like `simulate = "generate_waveforms.py:main"`. This will allow you to run the command `poetry run simulate` and have it execute that function for you.
- If you're going to run this project as a script, which I presume is the intention, you'll need a way to pass arguments from the command line to the `main` function. That `hermes.typeo` library wrote is meant to do exactly that: parse a function definition for the relevant arguments and then expose them to the command line. If you include the line `"hermes.typeo" = "^1.4"` in your dependencies section, you can add `from hermes.typeo import typeo` to the top of your script and then decorate main with `@typeo`, and that will take care of exposing all those arguments at the command line for you (so you could run e.g. `poetry run simulate -h` and it will tell you all the arguments that `main` is expecting.
- For tests, use [pytest](https://docs.pytest.org/en/7.0.x/) and include it as a dev dependency like how I've set it up [here](https://github.com/alecgunny/BBHNet/blob/d2353d6a3841b2ad08ebcdbfb6a64793ddbea4b6/libs/data/pyproject.toml#L20-L21). The `tests` directory should live at the same level as `pyproject.toml`, and include a file for each file in your library/project (in this case, just 1). They should import the actual code (`from generate_waveforms import main`), and execute it in functions that begin with `test_`.
- Include any files that you need for testing (e.g. the prior files) in the directory itself (somewhere like `tests/prior_files`) and track it with git (i.e. add them to the repo), this way anyone can run your tests and get the same results.
- If your tests involve writing data, consider using pytest [fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) for creating temporary directories then deleting them at the end of tests (you can find an example of something like this [here](https://github.com/alecgunny/BBHNet/blob/d2353d6a3841b2ad08ebcdbfb6a64793ddbea4b6/libs/data/tests/test_dataloader.py#L15-L20)).
- Minor point, but consider adopting [Google docstring style](https://google.github.io/styleguide/pyguide.html#381-docstrings), specifically using `Args:` instead of `Arguments:` and indenting individual argument descriptions rather than using dashes. This might seem trivial, but eventually we'll want to use libraries like Sphinx for writing documentation automatically from docstrings, so adopting a project-wide standard will help make that easier, and this is the convention I've programmed `typeo` to use to be able to automatically parse your docstring and display it as part of the command-line help.

wbenoit26:
> Can any of the functionality in simulate.py use any of the functions you wrote for libs/injections?
Not without modifying the injection functions as they're currently written. That could be a smart thing to do, though. I'll look into it.

> consider just having a single file called generate_waveforms.py in the same directory as your pyproject.toml
Done

> Consider adding your script as a command line executable in your poetry environment
Done

> include the line "hermes.typeo" = "^1.4" in your dependencies section...
Done

> Include any files that you need for testing (e.g. the prior files) in the directory itself 
Done

> consider adopting Google docstring style
Done

Everything regarding pytest is not done, except that I added it as a dependency.

wbenoit26:
I've split up a function in injection.py so that there are separate methods for generating raw waveforms and interferometer projection. This allows to use the raw waveform generation function in the dataset generation script.

I've also put my testing into what I think is the proper pytest format.

alecgunny:
This all looks really good. I'll pull everything and try to run it sometime in the next couple days and then we should be good to go. I'll also add in a CI unit testing workflow like the one I added into #21 to run these tests automatically before any future PR can be accepted

alecgunny:
I had to add `lalsuite` as a dependency of the `injections` library and constrain the `generate_waveforms` project to use `astopy<5.0` since versions `>=5.0` have issues with `gwpy`, but `bilby` tries to take the latest version it can get. But after that and a minor path issue in one of the fixtures everything passed! I'll add a GH workflow for running the tests automatically tomorrow AM and then we'll merge it in

libs/injection/bbhnet/injection/injection.py
```diff
@@ -92,12 +136,16 @@ def generate_gw(
         psi = p["psi"]
 
         # Generate signal in IFO
-        polarizations = waveform_generator.time_domain_strain(p)
         signal = np.zeros(waveform_size)
         for mode, polarization in polarizations.items():
             # Get ifo response
             response = ifo.antenna_response(ra, dec, geocent_time, psi, mode)
-            signal += response * sig.filtfilt(b, a, polarization)
+            signal += (
```
alecgunny, line 143: this is kind of a kooky way for black to factor this. Is there a reason it can't be `signal += response * polarization[i]` all on one line?

alecgunny (approved):
Just one small style comment but otherwise looks really good, don't want to hold us up over that

libs/injection/bbhnet/injection/injection.py
wbenoit26: Pretty sure that's a result of one the pre-commit hooks editing things. I don't recall writing it out that way...

alecgunny: It's all good I had to add another commit in anyway to fix a different issue with Black so I just fixed this quickly

Issue #11, opened 2022-03-08
Build random sampling data loader for glitches and waveforms


alecgunny:
Build a data loader for training that can sample from a dataset of glitches (#9 ) and dataset of waveforms (#10 ) as well as randomly time-slid background

alecgunny:
As a very rough blueprint, I can imagine building something like the following, which will be fully random so it's sort of up to you to determine how long you want an "epoch" to be (i.e. it doesn't move sequentially through the data). It's way less than optimal (way too much looping, but not quite sure how to efficiently sample kernels from a 1D timeseries), and is missing the injection portion, but hopefully this should be somewhere to start a discussion.

```python
from typing import Optional

import h5py
import numpy as np
import torch


class RandomWaveformDataset:
    def __init__(
        self,
        hanford_background: str,
        livingston_background: str,
        kernel_length: float,
        sample_rate: float,
        batch_size: int,
        batches_per_epoch: int,
        waveform_dataset: Optional[str] = None,
        waveform_frac: float = 0,
        glitch_dataset: Optional[str] = None,
        glitch_frac: float = 0,
        device: torch.device = "cuda"
    ) -> None:
        # sanity check our fractions
        assert 0 <= waveform_frac <= 1
        assert 0 <= glitch_frac <= 1

        # this one actually won't be sufficient:
        # e.g. if they add up to 0.99 but our batch
        # size is 32, they still won't leave any room
        # for background
        assert (waveform_frac + glitch_frac) <= 1

        # load in the background data
        # TODO: maybe these are gwf and we resample?
        with h5py.File(hanford_background, "r") as f:
            self.hanford_background = f["whatever"][:]

            # move everything onto the GPU up front so that
            # we don't have to pay for transfer time later.
            # If our datasets are on the scale of ~GBs this
            # shouldn't be a problem, esp. for the current
            # size of BBHNet. For the rest of these I'll
            # just do this in one fell swoop
            self.hanford_background = torch.Tensor(
                self.hanford_background
            ).to(device)
        with h5py.File(livingston_background, "r") as f:
            self.livingston_background = torch.Tensor(
                f["whatever"][:]
            ).to(device)

        # ensure that we have the same amount of
        # data from both detectors
        assert len(self.hanford_background) == len(self.livingston_background)

        # load in any waveforms if we specified them
        # TODO: what will the actual field name be?
        # TODO: will these need to be resampled?
        if waveform_dataset is not None:
            assert waveform_frac > 0
[...]

PR #21, opened 2022-03-10
Adding iterable dataset for training

alecgunny:
Addressing #11


alecgunny:
- [ ] Build tests for `glitch_frac > 0`
- [ ] Build tests for `waveform_frac > 0`
- [ ] Sample start_indices from first half of arrays to always include trigger
- [ ] Build CI for running tests on PR

libs/data/bbhnet/data/dataloader.py
```diff
@@ -0,0 +1,449 @@
...
+    def _sample_geocent_time(self, num_samples):
+        raise NotImplementedError
+
+    def inject_waveforms(self, background, waveforms):
+        sample_params = {
+            "ra": self._sample_ra(len(background)),
+            "dec": self._sample_dec(len(background)),
+            "psi": self._sample_psi(len(background)),
+            "geocent_time": self._sample_geocent_time(len(background)),
+        }
```
alecgunny, line 353: @wbenoit26 any ideas as to what these sampling functions might be? I assume at least one of them is like a uniform [0, pi] sample or something but not familiar enough with the meaning of all the parameters to be sure

libs/data/bbhnet/data/dataloader.py
```diff
@@ -0,0 +1,449 @@
...
+        for i, ifo in enumerate(["h1", "l1"]):
+            signal = project_raw_gw(
+                waveforms,
+                sample_params,
+                self.waveform_generator,
+                ifo,
+                get_snr=False,
+            )
+            signals[:, i] = signal
+        return background + torch.Tensor(signals).to(self.device)
```
alecgunny: @wbenoit26 is something like this sufficient? Or will I need to do the SNR reweighting you do [here](https://github.com/ML4GW/BBHNet/blob/9564f1595a13fa243ebc90a97b7c4025c2f6b3b6/libs/injection/bbhnet/injection/injection.py#L295-L300)? Does the fact that I do whitening by the background PSD help things at all?

Note that at this point the waveforms have been windowed into some `kernel_size` window around the trigger time. Will this affect the ifo response calculation?

alecgunny:
Moving from draft to full PR so that we can start review comments on individual lines. Injection code still needs to be completed and needs tests (though not sure what those will look like), but otherwise should be most of the way there

libs/data/bbhnet/data/dataloader.py
wbenoit26: We have some freedom to do what we want here, but I think it makes sense to stick with the priors from some of bilby's example prior dictionaries. They have a Uniform(0, 2*pi) prior on RA, a Cosine prior on Dec, and if the black hole is spinning, a Uniform(0, pi) prior on psi (0 if not spinning). We could have bilby do the sampling here too, but it may not be worth the overhead.

Geocent_time has just been a uniform on the duration of the frame, plus the initial time. For my injection code, there's some logic to prevent signals from overlapping.

libs/data/bbhnet/data/dataloader.py
wbenoit26: The SNR scaling will need to happen here as well if we want to have control over the range we're training on, and it might be a good idea to store the SNR regardless. Whitening will change the SNR, but I don't think that matters if all we're looking for is some consistent way to tune signal strength.

That's a good question. I think the answer is no, but I can do some digging to find out if that's true.

wbenoit26: Based on bilby's source code, the windowing will not affect the ifo response - it's a purely geometric quantity.

One thing that I do want to flag is the [get_detector_response](https://lscsoft.docs.ligo.org/bilby/_modules/bilby/gw/detector/interferometer.html#Interferometer.get_detector_response) function here, which I suppose is what Tri based his original injection code on. Presumably he had a reason that he needed to customize this code rather than just calling the function, but that might not be the case for us anymore.

We have a working injection script, so I don't want to keep iterating over it, but I did want to note that it exists.

alecgunny: Sounds good, I'll try to get all these fixes in today so we can merge ASAP.

Good catch on the built-in function. Agreed that since what we have works we may as well keep it, but worth keeping in mind down the line.

libs/data/bbhnet/data/dataloader.py
alecgunny: @wbenoit26 I opted to just use bilby's samplers since they were the most straightforward to implement. I may experiment with using scipy's distribution functions (since numpy doesn't support cosine), but this should work for now.

libs/data/bbhnet/data/dataloader.py
alecgunny: SNR reweighting is implemented, and the windowing ends up being a non-factor because I use the whole signal so that the fft calculation during SNR reweighting is more stable, then window it after it's been projected and reweighted

alecgunny:
@wbenoit26 I think everything should be in here by now, with tests for everything as well. The waveform sampling tests could be a bit more complete, but it's worth having a discussion at some point about what functionality we want to trust that `bbhnet.injection` has tested for us, and which we want to verify in downstream libs. In either case, I've noted areas of improvement with TODO's and can open these up as good-first-issues.

libs/data/bbhnet/data/dataloader.py
```diff
@@ -0,0 +1,372 @@
...
+        assert 0 <= waveform_frac <= 1
+        assert 0 <= glitch_frac <= 1
+
+        self.sample_rate = sample_rate
+        self.kernel_size = int(kernel_length * sample_rate)
+        self.batch_size = batch_size
+        self.batches_per_epoch = batches_per_epoch
+        self.device = device
+
+        # load in the background data and build time-domain
```
EthanMarx, line 182: I assume these background files will contain real data. How many seconds of data are we using? Are we using all of this data to calculate the asd? Are we using the same asd to whiten all samples? 

Not super relevant to code, but important questions I've had that I figured should document somewhere.

EthanMarx, line 182: I assume these background files will contain real data. How many seconds of data are we using? Are we using all of this data to calculate the asd? Are we using the same asd to whiten all samples? 

Not super relevant to code, but important questions I've had that I figured should document somewhere.

libs/data/bbhnet/data/dataloader.py
```diff
@@ -0,0 +1,372 @@
...
+        # create a target tensor, marking all the glitch data as 0.
+        y = torch.zeros((self.batch_size,))
+
+        # replace some of this data with glitches if
+        # we have glitch data to use
+        if self.glitch_sampler is not None:
+            hanford_glitches, livingston_glitches = self.glitch_sampler.sample(
+                self.num_glitches, self.kernel_size
+            )
+
```
EthanMarx, line 347: To clarify, this portion is creating "incoherent" glitches. i.e., no data-sample has a glitch in both Hanford and Livingston. 

In the future, might be worth considering adding some fraction of samples w/ glitches in both Hanford and Livingston.

EthanMarx, line 347: To clarify, this portion is creating "incoherent" glitches. i.e., no data-sample has a glitch in both Hanford and Livingston. 

In the future, might be worth considering adding some fraction of samples w/ glitches in both Hanford and Livingston.

libs/data/bbhnet/data/utils.py
```diff
@@ -0,0 +1,94 @@
...
+        raise ValueError(
+            "Must specify number of samples N when x is 1 dimensional"
+        )
+    elif len(x.shape) == 1:
+        # this is a 1D array, so we'll just be sampling
+        # the N kernels from the existing timeseries
+        if N > (len(x) - size):
+            raise ValueError(
+                "Array of length {} too short to sample {} "
+                "kernels of length {}".format(len(x), size, N)
```
EthanMarx, line 42: Should the N and size be switched in ValueError message: `.format(len(x), N, size)`

EthanMarx, line 42: Should the N and size be switched in ValueError message: `.format(len(x), N, size)`

libs/data/bbhnet/data/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+
+    if len(x.shape) == 1 and N is None:
+        raise ValueError(
+            "Must specify number of samples N when x is 1 dimensional"
+        )
+    elif len(x.shape) == 1:
+        # this is a 1D array, so we'll just be sampling
+        # the N kernels from the existing timeseries
+        if N > (len(x) - size):
```
EthanMarx, line 39: So for example, if `N = (len(x) - size)`, then the kernels will be made by shifting over one sample?

EthanMarx, line 39: So for example, if `N = (len(x) - size)`, then the kernels will be made by shifting over one sample?

libs/data/bbhnet/data/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    for i in idx:
+        # in the 1D case, the idx represent start samples
+        # from the timeseries
+        if len(x.shape) == 1:
+            samples.append(x[i : i + size])
+            continue
+
+        # otherwise, i represents an index along x's 0th dimension,
+        # and we sample from within a kernel's length of
+        # the center of each sampled index, where it is assumed
```
EthanMarx, line 77: i.e.  For signals, the "peak" or "coalescence" can lie anywhere in the kernel?

EthanMarx, line 77: i.e.  For signals, the "peak" or "coalescence" can lie anywhere in the kernel?

EthanMarx (approved):
I made a few comments, but mostly left some clarifying questions. Looks really nice excited to see this in action.

libs/data/bbhnet/data/dataloader.py
```diff
@@ -0,0 +1,372 @@
...
+        self.device = device
+
+        # load in the background data and build time-domain
+        # filters for whitening using them
+        self.hanford_background, hanford_asd, t0 = _load_background(
+            hanford_background, sample_rate, device
+        )
+        self.livingston_background, livingston_asd, t0 = _load_background(
+            livingston_background, sample_rate, device
+        )
```
EthanMarx, line 189: I assume these background files will contain real data. How many seconds of data are we using? Are we using all of this data to calculate the asd? Are we using the same asd to whiten all samples? 

Not super relevant to code, but important questions I've had that I figured should document somewhere.

libs/data/bbhnet/data/dataloader.py
```diff
@@ -0,0 +1,372 @@
...
+
+            if hanford_glitches is not None:
+                X[: len(hanford_glitches), 0] = hanford_glitches
+                idx = len(hanford_glitches)
+            else:
+                idx = 0
+
+            if livingston_glitches is not None:
+                slc = slice(idx, idx + len(livingston_glitches))
+                X[slc, 1] = livingston_glitches
```
EthanMarx, line 356: To clarify, this portion is creating "incoherent" glitches. i.e., no data-sample has a glitch in both Hanford and Livingston.

In the future, might be worth considering adding some fraction of samples w/ glitches in both Hanford and Livingston.

libs/data/bbhnet/data/utils.py
EthanMarx, line 42: So for example, if `N = (len(x) - size)`, then the kernels will be made by shifting over one sample?

libs/data/bbhnet/data/utils.py
```diff
@@ -0,0 +1,94 @@
...
+        # that the "trigger" of the relevant event will live
+        if max_sample_start <= 0:
+            start = 0
+        else:
+            start = np.random.randint(min_sample_start, max_sample_start)
```
EthanMarx, line 82: i.e.  For signals, the "peak" or "coalescence" can lie anywhere in the kernel?

EthanMarx (approved):
I made a few comments, but mostly left some clarifying questions. Looks really nice excited to see this in action.

libs/data/bbhnet/data/dataloader.py
alecgunny: So I think ideally this implementation should be agnostic to those details, subject to the constraint that it's sufficiently small data to fit into (GPU) memory at once, so O(1GB). Which data and how much of it will change over time, but starting somewhere on the order of ~16K seconds might be ok to get started? Idk this is probably something good to discuss as part of #27 

But yeah the data will be real and all of it will be used to calculate the ASD, which will be used to whiten all samples both during training _and_ at inference time (the nice thing about structuring the whitening in Torch code is that we can save this step as part of the model so that it gets performed on the server at inference time).

libs/data/bbhnet/data/dataloader.py
alecgunny: Ah ok yeah this is a very good point. Would it make sense to merge as-is and make sure that we can get training to converge with this set up, but keep an issue open to fix this once we're confident we have things structured the right way?

libs/data/bbhnet/data/utils.py
alecgunny: Yes on both fronts, great catch. I'll fix that now

libs/data/bbhnet/data/utils.py
alecgunny: Yeah it can lie anywhere in the kernel, but it _has_ to end up in the kernel

libs/data/bbhnet/data/dataloader.py
EthanMarx: Got it. Just wanted to bring this up because the ASD is quite variable. Whitening an entire training set w/ the same asd may turn out to be problematic, although as a first pass this should be totally fine.

libs/data/bbhnet/data/dataloader.py
EthanMarx: Yes completely agree

EthanMarx (approved):

libs/data/bbhnet/data/dataloader.py
alecgunny: Yeah I think the idea in the context of DL is just to use this as a normalization step, and in that case you just want to make sure you're applying the same normalization to test-time data as you did during training so that the statistics stay the same. If the statistics of your input data have shifted so dramatically that this whitening is no longer helpful, that's a problem but it isn't necessarily specific to whitening: you just have to retrain your network because the distribution you optimized for is no longer the one you're sampling from.

We could do an ifo-wise 0-mean unit-variance standardization instead, but this might be more natural in this context? Idk definitely something to keep an eye on

PR #23, opened 2022-03-30
Fixed bug where the SNRs of the signals were not in the desired range

wbenoit26:
The intention was that `old_snr` should be a list of the geometric means of the SNRs of the same signal across all interferometers. I forgot the `axis` argument, so it was summing everything into a single value. This caused the SNRs to be scaled to the wrong values.

### DIFF
--- a/libs/injection/bbhnet/injection/injection.py
+++ b/libs/injection/bbhnet/injection/injection.py
@@ -14,7 +14,18 @@ from gwpy.timeseries import TimeSeries
 
 
 def calc_snr(data, noise_psd, fs, fmin=20):
-    """ Calculate the waveform SNR given the background noise PSD"""
+    """Calculate the waveform SNR given the background noise PSD
+
+    Args:
+        data: timeseries of the signal whose SNR is to be calculated
+        noise_psd: PSD of the background that the signal is in
+        fs: sampling frequency of the signal and background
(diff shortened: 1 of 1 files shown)


alecgunny:
Looks good! Are there any tests you could add to this library to help double check for the desired behavior?

wbenoit26:
Yes, with any luck I'll have those written tomorrow.

alecgunny (commented):
This all looks good to me. Do you want to add a step to the ci workflow like [this one](https://github.com/ML4GW/BBHNet/blob/1e20b814cc541a74410917a40c91668591196110/.github/workflows/ci-tests.yaml#L30-L33) so that these tests run automatically on PR and merge?

wbenoit26:
All done! Is it an issue if these tests take a while? Last I tried it, it took a little under 40 minutes for everything to pass

alecgunny:
The CI test is failing right now because of a limitation of the pinto tool I built. I'll push changes to this branch sometime today that should fix it and then we can merge it in

alecgunny:
Also the timing shouldn't be an issue. Ideally it would be a bit shorter than that, or at least stop after the first failure so that we can debug without having to wait, but that's a bridge we can cross if/when this becomes an issue

alecgunny (approved):

PR #39, opened 2022-04-26
Training script

EthanMarx:


EthanMarx:
@alecgunny @eric-moreno 

Right now, the[ training script](https://github.com/EthanMarx/BBHNet/blob/training-script/libs/trainer/bbhnet/trainer/trainer.py) takes in 'pre-loaded' RandomWaveformDatasets for training and validation. 
Where do we want to this 'pre-loading' (e.g. loading in hdf5 files, setting glitch fraction, batch_size, etc).
I see that DeepClean has a [training project](https://github.com/ML4GW/DeepClean/blob/main/projects/sandbox/train/train/train.py) that seems to handle this. Is this something we want to adopt, and if so, should it be included in this pull reqeust?

EthanMarx:
Waiting on implementation of architecture library to fully test wrapper functionality.

alecgunny (commented):
This all looks really good. My only comment would be around the point you bring up about the `dict` data arguments to the `train` function. I go back and forth on this a lot too, but my general gut is that `dict`s that are supposed to have specific fields are just prone to errors and typos and are generally nonintuitive (to someone reading the code trying to understand it, it's not immediately clear what fields this dict is supposed to have and what it's not, even though it's obviously in the function doc). Some alternatives that jump off the top of my head:
1. Make a `dataclasses.dataclass` class called `DataStruct` or something that has `hanford_background`, `livingston_background`, `signal`, and `glitch` attributes and expose these as as the `train_data`, `valid_data` arguments. It's functionally the same as a `dict`, but at least the structure is made explicit somewhere and can be linted, autodoc'd, autocompleted, etc.
2. At that point, since we're grouping these files together anyway, just construct the `RandomWaveformDataset` objects in the `train` project and accept any iterable of `(X, y)` pairs as the arguments for `train_data` and `valid_data` in the `train` function. The nice thing about this is that most of the dataset args don't get used elsewhere in training: `sample_rate`, `min_snr`, `max_snr`, `batch_size`, `batches_per_epoch`. This also seems fair since our dataloader is one way, but perhaps not the only way, one might be interested in sampling data for BBHNet to train on. This is as opposed to the DeepClean case, where (and maybe this is just lack of imagination) the sampling process follows pretty straightforwardly from the problem statement.
3. Break the train arguments out explicitly: `hanford_background`, `livingston_background`, `signal`, `glitches`, and then have a single validation argument `valid_data` that's either the current dictionary or a 4-length `tuple` that follows the order in which the train arguments are listed and defaults to `None` (for no validation). This at least makes the arguments explicit once, and then having a line like
```python
if valid_data is not None:
    val_hanford_background, val_livingston_background, val_signal, val_glitches = valid_data
    ... # do dataloader stuff here
```
would help make it clear in one place what the structure is. This is also consistent with how Keras' `model.fit` method accepts validation data (which is why I opt for it in DeepClean where there's only two arguments).

I _think_ my intuition says option 2 makes the most sense, but not leaning super hard against any of them (including the current one). Thoughts?

EthanMarx:
> This all looks really good. My only comment would be around the point you bring up about the `dict` data arguments to the `train` function. I go back and forth on this a lot too, but my general gut is that `dict`s that are supposed to have specific fields are just prone to errors and typos and are generally nonintuitive (to someone reading the code trying to understand it, it's not immediately clear what fields this dict is supposed to have and what it's not, even though it's obviously in the function doc). Some alternatives that jump off the top of my head:
> 
> 1. Make a `dataclasses.dataclass` class called `DataStruct` or something that has `hanford_background`, `livingston_background`, `signal`, and `glitch` attributes and expose these as as the `train_data`, `valid_data` arguments. It's functionally the same as a `dict`, but at least the structure is made explicit somewhere and can be linted, autodoc'd, autocompleted, etc.
> 2. At that point, since we're grouping these files together anyway, just construct the `RandomWaveformDataset` objects in the `train` project and accept any iterable of `(X, y)` pairs as the arguments for `train_data` and `valid_data` in the `train` function. The nice thing about this is that most of the dataset args don't get used elsewhere in training: `sample_rate`, `min_snr`, `max_snr`, `batch_size`, `batches_per_epoch`. This also seems fair since our dataloader is one way, but perhaps not the only way, one might be interested in sampling data for BBHNet to train on. This is as opposed to the DeepClean case, where (and maybe this is just lack of imagination) the sampling process follows pretty straightforwardly from the problem statement.
> 3. Break the train arguments out explicitly: `hanford_background`, `livingston_background`, `signal`, `glitches`, and then have a single validation argument `valid_data` that's either the current dictionary or a 4-length `tuple` that follows the order in which the train arguments are listed and defaults to `None` (for no validation). This at least makes the arguments explicit once, and then having a line like
> 
> ```python
> if valid_data is not None:
>     val_hanford_background, val_livingston_background, val_signal, val_glitches = valid_data
>     ... # do dataloader stuff here
> ```
> 
> would help make it clear in one place what the structure is. This is also consistent with how Keras' `model.fit` method accepts validation data (which is why I opt for it in DeepClean where there's only two arguments).
> 
> I _think_ my intuition says option 2 makes the most sense, but not leaning super hard against any of them (including the current one). Thoughts?



> This all looks really good. My only comment would be around the point you bring up about the `dict` data arguments to the `train` function. I go back and forth on this a lot too, but my general gut is that `dict`s that are supposed to have specific fields are just prone to errors and typos and are generally nonintuitive (to someone reading the code trying to understand it, it's not immediately clear what fields this dict is supposed to have and what it's not, even though it's obviously in the function doc). Some alternatives that jump off the top of my head:
> 
> 1. Make a `dataclasses.dataclass` class called `DataStruct` or something that has `hanford_background`, `livingston_background`, `signal`, and `glitch` attributes and expose these as as the `train_data`, `valid_data` arguments. It's functionally the same as a `dict`, but at least the structure is made explicit somewhere and can be linted, autodoc'd, autocompleted, etc.
> 2. At that point, since we're grouping these files together anyway, just construct the `RandomWaveformDataset` objects in the `train` project and accept any iterable of `(X, y)` pairs as the arguments for `train_data` and `valid_data` in the `train` function. The nice thing about this is that most of the dataset args don't get used elsewhere in training: `sample_rate`, `min_snr`, `max_snr`, `batch_size`, `batches_per_epoch`. This also seems fair since our dataloader is one way, but perhaps not the only way, one might be interested in sampling data for BBHNet to train on. This is as opposed to the DeepClean case, where (and maybe this is just lack of imagination) the sampling process follows pretty straightforwardly from the problem statement.
> 3. Break the train arguments out explicitly: `hanford_background`, `livingston_background`, `signal`, `glitches`, and then have a single validation argument `valid_data` that's either the current dictionary or a 4-length `tuple` that follows the order in which the train arguments are listed and defaults to `None` (for no validation). This at least makes the arguments explicit once, and then having a line like
> 
> ```python
> if valid_data is not None:
>     val_hanford_background, val_livingston_background, val_signal, val_glitches = valid_data
>     ... # do dataloader stuff here
> ```
> 
> would help make it clear in one place what the structure is. This is also consistent with how Keras' `model.fit` method accepts validation data (which is why I opt for it in DeepClean where there's only two arguments).
> 
> I _think_ my intuition says option 2 makes the most sense, but not leaning super hard against any of them (including the current one). Thoughts?

@alecgunny I think option 1 makes the most sense to me, but I think that may be because I'm a bit confused about option 2. What exactly are the (X, y) pairs you are talking about ? I would expect if we're creating the RandomWaveformDatasets in the `train` project, that we would explicitly have the `train` function accept RandomWaveformDataset objects as well. Is your idea that in the future, its possible we may use the train function w/o RandomWaveformDataset objects?

I see your point about how the `train` function is agnostic to the parameters `sample_rate`, `min_snr`, `max_snr`, `batch_size`, `batches_per_epoch`.

alecgunny (approved):
Looks good to me. Couple issues we can open up to improve things once it's merged but I think it's worth just getting in, especially if training is running. But I'll note them here for reference.
- Make mixed precision optional and default to being off ([DeepClean](https://github.com/ML4GW/DeepClean/blob/main/libs/trainer/deepclean/trainer/trainer.py) now has an example of this)
- Make validation data optional in train script/fn
- Now that we've implemented the arch/training wrappers in a new lib, as well as the single training step, start thinking about what we can generalize into some sort of training utils library with its own repo in ML4GW

PR #42, opened 2022-05-10
Architecture zoo

alecgunny:
Adding a library for model architectures initialized with a simple 1D resnet implementation. Implementation is largely lifted from [`torchvision`'s implementation](https://github.com/pytorch/vision/blob/main/torchvision/models/resnet.py), with a few changes for supporting 1D convolutions as well as arbitrary kernel sizes and depths given the difference of our problem from the usual CV ones this arch was built for.


alecgunny:
Right now wrapper tests are failing because pinto can't install extras (ml4gw/pinto#7). This should be a reasonably quick fix on the pinto side, so I'll implement this this afternoon and then update the workflow here to use this feature.

EthanMarx (approved):

EthanMarx (approved):

PR #44, opened 2022-05-11
generalize dataloader for arbitrary ifos

EthanMarx:


EthanMarx:
Now being done in #139

PR #47, opened 2022-05-14
incorporate omicron into glitch generation project

EthanMarx:
fully Integrate omicron via pyomicron library into the glitch-generation project so a user does not need to rely on a previously made omicron triggers.


projects/generate_glitches/generate_glitches.py
```diff
@@ -103,38 +106,35 @@ def generate_glitch_dataset(
...
-    triggers = np.loadtxt(trig_file)
+    triggers = h5py.File(trig_file)["triggers"][()]
 
     # restrict triggers to within gps start and stop times
     # and apply snr threshold
-    times = triggers[:, time_col]
+    times = triggers["time"][()]
     mask = (times > start) & (times < stop)
-    mask &= triggers[:, snr_col] > snr_thresh
+    mask &= triggers["snr"][()] > snr_thresh
```
alecgunny, line 127: Does it make sense to do these in an `with h5py.File(trig_file, "r") as f:` context just in case anything goes wrong? Or even just grab the relevant `triggers`, `times`, and `snrs` arrays under the context and then do all the manipulation after they're loaded and the file is closed?

alecgunny (approved):
This is extremely slick, love it. Does omicron require any special credentials or anything? Other than that and my one very minor comment, all good.

EthanMarx:
> This is extremely slick, love it. Does omicron require any special credentials or anything? Other than that and my one very minor comment, all good.

PR #50, opened 2022-05-23
Constraining broken dependencies

### LINKED ISSUE #40: Create base lib for enforcing maximum versions
Several of the libraries required by BBHNet have strict maximum versions due mainly to inconsistencies with gwpy. The ones that occur to me off the to of my head are `astropy<5.0` and `scipy<1.8`.

Rather than having to add these restrictions to every library or project that might use them, it might make sense to create a dummy `bbhnet.base` project that is a dependency of all other libraries that does nothing but has these constraints in its `dependencies` table. This way, all other libraries will automatically inherit these restrictions, and any others we might need to enforce down the line.

alecgunny:
Adding dummy `base` library for constraining known issues with newer libraries. Fixes #40 and an issue with recreating the `generate_waveforms` project environment

### DIFF
--- a/libs/base/poetry.lock
+++ b/libs/base/poetry.lock
@@ -1,0 +1,194 @@
+[[package]]
+name = "astropy"
+version = "4.3.1"
+description = "Astronomy and astrophysics core library"
+category = "main"
+optional = true
+python-versions = ">=3.7"
+
+[package.dependencies]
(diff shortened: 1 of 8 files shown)


EthanMarx (approved):
Not much to say looks good. What exactly is the problem between gwpy and astropy / scipy that requires these version constraints?

PR #51, opened 2022-05-26
add tool.typeo table to generate waveforms pyproject

EthanMarx:
Update generate waveforms pyproject

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 **/__pycache__
 **/.ipynb_checkpoints
 **/*.log
-**/*.ipynb
 **/*.env
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 **/__pycache__
 **/.ipynb_checkpoints
+**/*.log
+**/*.ipynb
+**/*.env
--- a/projects/generate_waveforms/pyproject.toml
+++ b/projects/generate_waveforms/pyproject.toml
@@ -16,7 +16,7 @@ pytest = "^6.2"
 [tool.typeo]
 outdir = "${OUT_DIR}"
 n_samples = 10000
-prior_file = "${PRIOR_FILE}"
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,2 +1,5 @@
 **/__pycache__
 **/.ipynb_checkpoints
+**/*.log
+**/*.ipynb
```
alecgunny, line 4: I'm not sure we want to exclude all notebooks do we? I can imagine we may eventually want to use them for viz or blog post-type purposes. If you have some in your local directory you'd like to ignore, you might want to consider using the `.git/info/exclude` file as described [here](https://stackoverflow.com/questions/1753070/how-do-i-configure-git-to-ignore-some-files-locally)

projects/generate_waveforms/pyproject.toml
```diff
@@ -13,6 +13,16 @@ python = ">=3.9,<3.11"
 [tool.poetry.dev-dependencies]
 pytest = "^6.2"
 
+[tool.typeo]
+outdir = "${OUT_DIR}"
+n_samples = 10000
+prior_file = "${PRIOR_FILE}"
```
alecgunny, line 19: Is there a specific setting of priors that's worth checking in to the repo so that we can version it alongside everything else? These prior files are pretty small right?

alecgunny (changes requested):
Only major thing for me is the `.gitignore`, otherwise looks good

.gitignore
EthanMarx: Yeah this was a local problem. Thanks for the pointer

projects/generate_waveforms/pyproject.toml
EthanMarx: Yes that's a good idea. I'll put one together

alecgunny (approved):

PR #57, opened 2022-05-26
Generate background project

EthanMarx:
create project that generates background dataset

### DIFF
--- a/projects/generate_background/generate_background.py
+++ b/projects/generate_background/generate_background.py
@@ -84,3 +84,4 @@ def main(
 
         with h5py.File(outdir / f"{ifo}_background.h5", "a") as f:
             f.create_dataset("hoft", data=data)
+            f.create_dataset("t0", data=float(segment[0]))
--- a/projects/generate_background/generate_background.py
+++ b/projects/generate_background/generate_background.py
@@ -62,7 +62,7 @@ def main(
 
     logging.info(
         "Querying coincident, continuous segment "
-        "from {segment[0]} to {segment[1]}"
+        "from {} to {}".format(*segment)
     )
 
     for ifo in ifos:
--- a/projects/generate_background/generate_background.py
+++ b/projects/generate_background/generate_background.py
@@ -1,0 +1,86 @@
+import logging
+from pathlib import Path
+from typing import List
+
+import h5py
+import numpy as np
+from gwdatafind import find_urls
+from gwpy.segments import DataQualityDict
+from gwpy.timeseries import TimeSeries
+from hermes.typeo import typeo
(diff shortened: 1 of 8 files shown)


projects/generate_background/generate_background.py
```diff
@@ -0,0 +1,87 @@
...
+            urltype="file",
+        )
+        data = TimeSeries.read(
+            files, channel=f"{ifo}:{channel}", start=segment[0], end=segment[1]
+        )
+
+        # resample
+        data = data.resample(sample_rate)
+
+        with h5py.File(outdir / f"{ifo}_background.h5", "a") as f:
```
alecgunny, line 85: Is there a reason to use `"a"` here instead of `"w"`?

projects/generate_background/generate_background.py
```diff
@@ -0,0 +1,87 @@
...
+        raise ValueError(
+            "No segments of minimum length, not producing background"
+        )
+
+    # choose first of such segments
+    segment = intersection[continuous_segments[0]]
+
+    logging.info(
+        "Querying coincident, continuous segment "
+        "from {segment[0]} to {segment[1]}"
```
alecgunny, line 65: no `f` used here, won't print correctly. Alternatively, if `segment` has just the 2 elements, maybe consider `"from {} to {}".format(*segment)`

projects/generate_background/generate_background.py
```diff
@@ -0,0 +1,87 @@
...
+    Args:
+        start: start gpstime
+        stop: stop gpstime
+        ifos: which ifos to query data for
+        outdir: where to store data
+    """
+
+    # query segments for each ifo
+    # I think a certificate is needed for this
+    segments = DataQualityDict.query_dqsegdb(
```
alecgunny, line 36: I think you're right. This should probably be fine for now if we're just running this from LDG. Longer term if we want to run unit tests from a GitHub runner we it looks like the underlying call `gwpy` makes (through `dqsegdb`) to get credentials is [this](https://gwdatafind.readthedocs.io/en/stable/api/gwdatafind.utils.find_credential.html). So we can think about adding [Kerberos credentials](https://computing.docs.ligo.org/guide/auth/#ligo-kerberos) as a secret to the GitHub workflow then use it to generate an [X.509 credential](https://computing.docs.ligo.org/guide/auth/#ligo-x509) at a location saved to the appropriate environment variable. Not that we need any of this right now, just documenting it here for posterity.

alecgunny (approved):
Just a couple quick points that don't affect the overall quality of the PR, so approving it so you can merge ASAP

PR #58, opened 2022-05-27
implement data generation pipeline via pinto

EthanMarx:
implement data generation pipeline via pinto


EthanMarx:
Was able to get the pipeline running end to end with a converging training loss!

I might have incorporated a few too many things into this one pull request. I'll try to summarize below:

1. Created pinto pipeline connecting all data generation projects and training
2. Implemented optional validation into the trainer library, as well as the train project
3. Incorporated boolean for turning on and off mixed precision training

libs/trainer/bbhnet/trainer/trainer.py
```diff
@@ -271,4 +283,23 @@ def train(
                     )
                     break
 
+        # if no validation dataset passed, save model based on
+        # best training loss. i.e. attempt to overfit
+        else:
```
alecgunny, line 288: Not entirely sure this is necessary: for any sufficiently complex model, your train loss will almost always be able to approach 0 arbitrarily. But it's probably fine to keep in here for now

libs/trainer/bbhnet/trainer/trainer.py
```diff
@@ -271,4 +283,23 @@ def train(
...
+                logging.debug(
+                    "No validation dataset passed. "
+                    "Achieved new lowest training loss, "
+                    "saving model weights "
+                )
+                best_train_loss = train_loss
+                weights_path = os.path.join(outdir, "weights.pt")
+                torch.save(model.state_dict(), weights_path)
+
+    print(history)
```
alecgunny, line 304: Is this a leftover debugging artifact or do you want this for the purposes of reaching pinto's stdout?

libs/trainer/bbhnet/trainer/trainer.py
alecgunny: Yeah true actually this is probably important for LR scheduling since we're not doing it based on epoch: as you approach a minimum you'll want to take smaller steps to avoid stepping right over it and sort of "circling the drain". Normally if you use like a cosine or exponential decay this isn't an issue since the LR decreases on its own, but since we use a plateau LR scheduler you'll still need to manage this manually.

I guess more broadly this is probably less important just because it will generally be the case that if you're not doing validation, this probably isn't a run you're doing to achieve good results but to sanity check your training code, and in that sense this probably doesn't matter too much. But either way it's definitely fine to include for now since it doesn't negatively impact anything

PR #61, opened 2022-06-02
IaaS export project

alecgunny:
- [x] add `StandardScaler` subclass of `bbhnet.data.transforms.transform.Transform`
- [x] unit tests for `bbhnet.data.transforms` submodule
- [x] change unit tests for existing dataloader to get rid of whitening
- [x] update training tests for new wrapper behavior
- [x] fill out `export` project
- [x] unit tests for `export` project, add to CI


alecgunny:
New changes implement whitening as a separate preprocessing `torch.nn.Module` that gets fit to the training data before getting passed in. This should obviate the need to "fit" the validation data loader to the training data's background: everything gets run through the whitening module before it gets to the model. @eric-moreno not sure how this impacts your PR, but just wanted to put it on your radar. Also up for discussing whether this is the best way of doing things.

alecgunny:
Issues to open on close
- Test `architecturize` behavior when called with `architecture` args or kwargs

libs/data/bbhnet/data/transforms/whitening.py
```diff
@@ -0,0 +1,104 @@
...
+                fftlength=self.fftlength, window="hanning", method="median"
+            )
+            asd = asd.interpolate(1 / self.kernel_length).value
+            tdf = fir_from_transfer(
+                1 / asd, ntaps=ntaps, window="hanning", ncorner=0
+            )
+            tdfs.append(tdf)
+
+        tdf = np.stack(tdfs)[:, None, :-1]
+        self.set_value(self.time_domain_filter, tdf)
```
EthanMarx, line 80: Where exactly was the whitening step going wrong? I assume it was one of these odd dimension transformations that are hard to follow... 

Just making sure we found the logic error and didn't just get lucky.

libs/trainer/bbhnet/trainer/trainer.py
```diff
@@ -179,6 +177,13 @@ def train(
     model = architecture(num_ifos)
     model.to(device)
 
+    # if we passed a module for preprocessing,
+    # include it in the model so that the weights
+    # get exported along with everything else
+    if preprocessor is not None:
+        preprocessor.to(device)
+        model = torch.nn.Sequential(preprocessor, model)
+
```
EthanMarx, line 186: So if I understand this correctly, for the StandardScaler, the standardization will be done batch wise, not over the entire training set.

projects/sandbox/export/export.py
```diff
@@ -0,0 +1,207 @@
...
+    sample_rate: float,
+    weights: Optional[Path] = None,
+    streams_per_gpu: int = 1,
+    instances: Optional[int] = None,
+    platform: qv.Platform = qv.Platform.ONNX,
+    clean: bool = False,
+    verbose: bool = False,
+) -> None:
+    """
+    Export a DeepClean architecture to model repository
```
EthanMarx, line 30: Replace DeepClean w/ BBHnet in doc string

projects/sandbox/train/train.py
```diff
@@ -115,6 +118,18 @@ def main(
...
+    # we just expose this as an arg? How will this fit in
+    # to the broader-generalization scheme?
+    preprocessor = WhiteningTransform(2, sample_rate, kernel_length)
+
+    # TODO: make this a `train_dataset.background` `@property`?
+    background = torch.stack(
+        [train_dataset.hanford_background, train_dataset.livingston_background]
+    )
+    preprocessor.fit(background)
+
```
EthanMarx, line 132: Add preprocessor type as argument so we can use StandardScaler?

libs/data/bbhnet/data/transforms/whitening.py
EthanMarx: Hmm thats a bit worrisome. At least now that the whitening transform is its own module I can play around with it separately and attempt to isolate the issue.

libs/trainer/bbhnet/trainer/trainer.py
alecgunny: Yeah so all these preprocessing modules are meant to be called batch-wise so that they can be built into the model and their parameters get exported with it. Given the fact that some of our data gets created at loading time, it makes less sense to preprocess everything up front

projects/sandbox/train/train.py
alecgunny: So I was thinking about something like this, but preprocessors will have their own keyword arguments (e.g. `StandardScaler` doesn't have any args besides `num_ifos`). This might invite a wrapper like `architecturize`, but this starts to get complicated pretty fast and since in principle the preprocessing method shouldn't change with anywhere near the frequency of the architecture (there should be a "right" way to do preprocessing for a particular use case), I don't think it's too much to ask individual projects to build their preprocessing methods explicitly

PR #63, opened 2022-06-03
CI update for new pinto release

alecgunny:
An update of to `pinto` changed the way project paths get handled. This PR updates the CI workflow to use the new syntax, and introduces some small path-manipulation changes in the `injection` library so that the workflow actually gets run, since so many libraries and projects depend on it.


EthanMarx (approved):
thumbs up.

PR #67, opened 2022-06-10
H5 tests

rafia17:
I have changed two files:

h5.py: added the checks to make sure that the length of y and dataset arrays matches the timeseries array. Raise a ValueError if it doesn't

test_h5.py: added tests to check if the ValueError is raised appropriately


libs/io/bbhnet/io/h5.py
```diff
@@ -0,0 +1,55 @@
+from typing import TYPE_CHECKING, Optional, Tuple
+
+import h5py
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+import numpy as np
```
alecgunny, line 8: This should be under the TYPE_CHECKING `if` statement, correct?

libs/io/bbhnet/io/h5.py
```diff
@@ -0,0 +1,55 @@
...
+    with h5py.File(fname, "w") as f:
+        f["GPSstart"] = t
+        if y is not None:
+            if len(y) != len(t):
+                raise ValueError("Length of y and t doesn't match")
+            f["out"] = y
+
+        for key, value in datasets.items():
+            if len(value) != len(t):
+                raise ValueError("Length of t and data arrays don't match")
```
alecgunny, line 42: Consider using the `key` in this error so its more explicit as to which array is the offender

libs/io/tests/test_h5.py
```diff
@@ -0,0 +1,171 @@
...
+        for key, value in datasets.items():
+            f[key] = value
+
+    # test the read_timeseries function
+    values = h5.read_timeseries(fname, *datasets)
+
+    # each array in "values", except the last one
+    #  should match with dataset arrays
+    for key, value in zip(datasets, values[:-1]):
+        assert (value == datasets[key][:]).all()
```
alecgunny, line 168: Small aesthetic point, but since `datasets[key]` is just an array, you actually don't need the `[:]` slice, which is just used as syntatic sugar for HDF5 `dataset` objects to extract the underlying data. It will still _work_ for arrays, but it's just saying "return a full slice of the array", which it would have done anyway.

alecgunny (changes requested):
The tests look good, just a couple minor points of feedback. It also looks like you currently have a merge conflict with the main branch, so you'll need to fix that before the CI tests can run.

libs/io/bbhnet/io/h5.py
rafia17: Done

libs/io/bbhnet/io/h5.py
```diff
@@ -0,0 +1,55 @@
...
+    y: Optional["np.ndarray"] = None,
+    **datasets,
+):
+    # read time information from time array, and map
+    # to ints for string formatting purposes if the
+    # times actually amount to ints
+    t0 = t[0]
+    t0 = int(t0) if int(t0) == t0 else t0
+
+    length = t[-1] - t[0]
```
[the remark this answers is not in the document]
rafia17: Thanks for explaining. I have made the change

libs/io/bbhnet/io/h5.py
rafia17: Added the "key" to the error message

libs/io/tests/test_h5.py
rafia17: removed the [:]

rafia17:
Hi Alec, I have incorporated the minor changes that you had requested and it needs your approval to merge.

libs/io/bbhnet/io/h5.py
```diff
@@ -39,7 +39,11 @@ def write_timeseries(
 
         for key, value in datasets.items():
             if len(value) != len(t):
-                raise ValueError("Length of t and data arrays don't match")
+                raise ValueError(
+                    "Length of data array:",
+                    key,
+                    "doesn't match the length of timeseries",
+                )
```
alecgunny, line 46: this will be rendered better as a single string: `f"Length of data array '{key}' doesn't match the length of timeseries"`

alecgunny (changes requested):
Looking great! One last small change and it's ready to merge (assuming the tests pass)

libs/io/bbhnet/io/h5.py
rafia17: This change is incorporated

alecgunny (approved):

PR #71, opened 2022-06-14
Refactor generate_gw; fixes #70

### LINKED ISSUE #70: Improvements to `bbhnet.injection.generate_gw`
- Accept `sample_rate` and `duration` arguments which are `Optional`. If `waveform_sampler` is not `None`, then these should be `None` (enforce this), otherwise use these arguments to instantiate a waveform sampler.
- Should `minimum_frequency` be an argument as well, with 20 as a default?
- How to dynamically infer the number of polarizations (to initialize the `signals` array) and ensure they are always concatenated in the same order?
- Can we do the filtering at the end once all the polarizations have been inserted into the `signals` array?

deepchatterjeeligo:


deepchatterjeeligo:
I also noticed that gwpy was missing from the dependecies, I added it

EthanMarx:
> I also noticed that gwpy was missing from the dependecies, I added it

@deepchatterjeeligo 
Actually gwpy is specified in the base repos `environment.yaml`. 
If you run `pinto build` inside of `libs/injection/`, pinto will recursively move up directories and look for an `environment.yaml,` install that via `conda`, and then ultimately `poetry install` the `libs/injection/` `project`

deepchatterjeeligo:
I'll admit, I have looked at this only for the last hour or so, but I think that the `generate_gw` can be refactored since it is hard to test. Should I refactor that as a part of this MR? Can you give me a summary a short summary of what this is doing @EthanMarx

deepchatterjeeligo:
Does the local dev setup require `pinto`? It seems like a production thing right?

EthanMarx:
Yes we use pinto for local development, since this allows us to develop/test the code in the environment it will ultimately be run in. The repo with installation instructions and documentation are [here](https://github.com/ML4GW/pinto)

libs/injection/bbhnet/injection/injection.py
```diff
@@ -39,15 +39,57 @@ def calc_snr(data, noise_psd, fs, fmin=20):
...
+
+    sampling_params = _set_missing_params(
+        default_waveform_sampling_params, waveform_generator_params
+    )
+    waveform_approximant_params = _set_missing_params(
+        default_waveform_approximant_params, waveform_generator_params
+    )
+
+    bilby_params = sampling_params.copy()
+    bilby_params.update(dict(waveform_arguments=waveform_approximant_params))
```
alecgunny, line 77: `sampling_params["waveform_arguments"] = waveform_approximant_params`

deepchatterjeeligo: done

libs/injection/bbhnet/injection/injection.py
```diff
@@ -39,15 +39,57 @@ def calc_snr(data, noise_psd, fs, fmin=20):
...
+
+    logging.info("Waveform parameters %s" % (bilby_params))
```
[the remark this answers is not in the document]
deepchatterjeeligo: done

deepchatterjeeligo:
OK. I think its good to go

alecgunny (approved):

PR #77, opened 2022-06-16
Improve injection tests

### LINKED ISSUE #69: Injection test parameterizations
Some of the parameters in the tests for `bbhnet.injections` have redundant values which are making these tests take a while. I'm also not sure we need to be fetching real open data for the strain backgrounds.

This is in addition to #55 , which deals with testing the _content_ of injections rather than just their shapes

deepchatterjeeligo:
mock `fetch_open_data` by constructing custom `TimeSeries`. Makes tests faster.

Fixes #69

### DIFF
--- a/libs/injection/tests/test_injection.py
+++ b/libs/injection/tests/test_injection.py
@@ -1,13 +1,13 @@
 #!/usr/bin/env python
 # coding: utf-8
-import os
-import shutil
+import tempfile
 from pathlib import Path
 from unittest.mock import patch
 
 import h5py
 import numpy as np
 import pytest
+from gwpy.detector import Channel
 from gwpy.timeseries import TimeSeries
 
 import bbhnet.injection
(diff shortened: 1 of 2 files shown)


deepchatterjeeligo:
Tests are failing with
```
E           RuntimeError: channel 'Strain' not found
```
So the mocking is not entirely correct. If anyone wanted to add more changes, feel free.

deepchatterjeeligo:
this is ready for review

libs/injection/tests/test_injection.py
```diff
@@ -142,22 +97,35 @@ def test_generate_gw(mock_filter):
...
 
 
+@pytest.mark.parametrize(
+    "ifos,n_samples,sample_rate,frame_duration,"
+    "waveform_duration,prior_filename",
+    [
+        (["H1"], 1, 4096, 2, 0.5, "nonspin_BBH.prior"),
+        (["H1", "L1"], 2, 2048, 20, 1, "precess_tides.prior"),
+    ],
+)
```
alecgunny, line 107: Is it worth testing cartesian products of these param combinations using fixtures, or is this sufficiently informative? Genuinely curious because I don't know enough about how the injections are created and what potential problem areas might be. Otherwise looks great.

deepchatterjeeligo: I have traditionally seen the use case of fixtures to be something like a mock object, something like a database client or something like that which is not accessible during testing. Here I thought the fixtures were used as a proxy for parameterization, and hence changed it to `pytest.mark.parametrize`. By cartesian product, do you mean all possible combinations? We could add those combinations to the parameterized tests.

alecgunny: Yeah so using parameterized fixtures will automatically run through a cartesian products of all parameter settings, but agreed this can run amok if applied lazily (e.g. you don't expect a change in one parameter to affect the outcome tested by a change in another parameter). But the issue here might be that if the first set of parameters works but the second fails, it's harder to localize exactly which parameter or parameters caused the failure.

deepchatterjeeligo: Sorry, for not being prompt with resolving this one. 

> But the issue here might be that if the first set of parameters works but the second fails, it's harder to localize exactly which parameter or parameters caused the failure.

Correct, using the `pytest.mark.parametrize` will show the parameters for which the tests pass or fail. Anyway, I added several more test cases, which include single vs double instruments, spanning long/short frame lengths, signal duration, waveform models. Hopefully this looks good.

alecgunny (approved):

Issue #81, opened 2022-06-17
`fir_from_transfer` `ntaps` keyword initialized improperly


EthanMarx:
@alecgunny 
I think this problem is the core of the whitening issue:

Currently the length of the convolutional window we use for whitening is `DEFAULT_FFTLENGTH * sample_rate `  (see [this line](https://github.com/ML4GW/BBHNet/blob/main/libs/data/bbhnet/data/transforms/whitening.py#L66)). However, since our `DEFAULT_FFTLENGTH = 2` and our data is 1s long, this means the convolutional window is longer than our data. The correct way to initialize `ntaps` is with the keyword argument `fduration` As noted [here](https://github.com/gwpy/gwpy/blob/b5f791105efc5f78ce51f35455bcde8a5d36cef3/gwpy/timeseries/timeseries.py#L1689#L1691) in the documentation for the gwpy whitening function. 

As noted [here](https://github.com/gwpy/gwpy/blob/b5f791105efc5f78ce51f35455bcde8a5d36cef3/gwpy/timeseries/timeseries.py#L1728#L1730), there is a 'filter set in' of `0.5 * fduration` of corrupted data. Thus in our case, our entire data segments are being corrupted. 

We can handle this by incorporating a `fduration` argument to the `WhiteningTransform` and fixing the `ntaps` initialization. However, we have to be careful since we'll have `0.5 * fduration` of data corrupted on both the front and back of the sample. If we set `fduration` sufficiently small, this probably won't be an issue. 

Alternatively, we hack the dataloader to pass `kernel_length + fduration` data to the model, and find a way to crop `0.5 * fduration` samples from the front and end of each sample. This is obviously more complicated. 

Let me know your thoughts. For now, I am going to implement the `fduration` keyword arg (or simply a `DEFAULT_FDURATION`) and initialize it to `0.01`.

EthanMarx:
Ultimately I think we need to solve this problem:
```
data = h5py.File(hanford_background)["hoft"][()]
ts = TimeSeries(data, dt = 1/ sample_rate)

whitened_then_cropped = ts.whiten(asd = asd).crop(2,3).value
cropped_then_whitened = ts.crop(2,3).whiten(asd = asd).value

```
Whitening a long segment then cropping to 1 second window produces good results, but
cropping the segment and then performing whitening gives garbage. There must be some combination of arguments
to the whiten function that will produce good behavior.

[...]

PR #82, opened 2022-06-19
Update whitening

EthanMarx:
This PR addresses issue #81. The `WhiteningTransform` object now takes in Optional parameters `highpass` and `fduration`.

### DIFF
--- a/libs/data/bbhnet/data/transforms/whitening.py
+++ b/libs/data/bbhnet/data/transforms/whitening.py
@@ -43,7 +43,7 @@ class WhiteningTransform(Transform):
         self.fduration = fduration or kernel_length / 2
 
         # initialize the parameter with 0s, then fill it out later
-        self.tdf_size = int(self.fduration * self.sample_rate)
+        self.ntaps = int(self.fduration * self.sample_rate)
 
         self.kernel_size = int(self.kernel_length * self.sample_rate)
 
@@ -52,9 +52,9 @@ class WhiteningTransform(Transform):
         # to confirm kernel_size is even first? Does that 0. still
         # get appended in the odd case?
         self.time_domain_filter = self.add_parameter(
-            torch.zeros((num_ifos, 1, self.tdf_size - 1)),
+            torch.zeros((num_ifos, 1, self.ntaps - 1)),
         )
-        self.window = torch.hann_window(self.tdf_size)
+        self.window = torch.hann_window(self.ntaps)
 
     def to(self, device: torch.device):
         """
@@ -84,7 +84,6 @@ class WhiteningTransform(Transform):
                 )
             )
 
-        ntaps = int(self.fduration * self.sample_rate)
         tdfs = []
         for x in X.cpu().numpy():
 
@@ -94,7 +93,10 @@ class WhiteningTransform(Transform):
             )
             asd = asd.interpolate(self.df).value
             tdf = fir_from_transfer(
-                1 / asd, ntaps=ntaps, window="hanning", ncorner=self.ncorner
+                1 / asd,
+                ntaps=self.ntaps,
+                window="hanning",
+                ncorner=self.ncorner,
             )
             tdfs.append(tdf)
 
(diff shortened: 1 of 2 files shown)


libs/data/bbhnet/data/transforms/whitening.py
```diff
@@ -88,17 +113,75 @@ def forward(self, X: torch.Tensor) -> torch.Tensor:
         X = X.transpose(2, 0)
         X = X - X.mean(axis=0)
         X = X.transpose(0, 2)
-        X *= self.window
+
+        self.pad = int(np.ceil(self.time_domain_filter.size(-1) / 2))
```
alecgunny, line 117: Probably worth either assigning as an attribute up front, or just using as a variable in this namespace (i.e. no `self.`, just `pad`)

libs/data/bbhnet/data/transforms/whitening.py
```diff
@@ -88,17 +113,75 @@ def forward(self, X: torch.Tensor) -> torch.Tensor:
...
+            conv[:, :, : nfft - self.pad] = torch.nn.functional.conv1d(
+                X[:, :, :nfft],
+                self.time_domain_filter,
+                groups=self.num_ifos,
+                padding="same",
+            )[:, :, : nfft - self.pad]
+
+            # process chunks of length nstep
+            k = nfft - self.pad
+            while k < self.kernel_size - nfft + self.pad:
```
alecgunny, line 163: Looping and doing multiple convolutions is going to be pretty inefficient on GPU, especially since each convolution will be comparatively small because it only operates on a subslice (basically you're going to waste time launching a bunch of kernels for ops that will be pretty small, while GPUs can parallelize individual operations really efficiently). Consider using the torch [unfold](https://pytorch.org/docs/stable/generated/torch.nn.Unfold.html) operation up front, which will basically create multiple windowed views of the timeseries at a given stride (which for reasons having to do with what this op is intended for, I believe you'll want to pass as the `dilation` argument, _not_ the `stride` argument).

Then you can do a single convolution along the last axis, slice off the pads so that the views are non-overlapping, then do a simple reshape. This solution might only work for timeseries whose length can be divided evenly into chunks like this, but it's probably worth enforcing this for efficiency, especially at inference time.

alecgunny (commented):
Looks really good, curious to hear your thoughts on a more efficient implementation. The response could even just be that we don't plan on using `fduration`s that are short enough to trigger this implementation for the time being, and so we don't need to worry about efficiency in this PR and can open it as an issue for later.

alecgunny:
Also the export tests take forever, partly because there's probably more permutations than is really necessary and partly because there's a bunch of permutations that are intended to fail (if the kernel length is less than the stride length) that still have to go to all the trouble of building and saving an NN that they will never use. If this is slowing you down, feel free to take away some of these permutations or I can also open up a quick PR to address the NN issue as well

libs/data/bbhnet/data/transforms/whitening.py
alecgunny: Yeah that sounds great

EthanMarx:
>Also the export tests take forever, partly because there's probably more permutations than is really necessary and partly because there's a bunch of permutations that are intended to fail (if the kernel length is less than the stride length) that still have to go to all the trouble of building and saving an NN that they will never use. If this is slowing you down, feel free to take away some of these permutations or I can also open up a quick PR to address the NN issue as well




> Also the export tests take forever, partly because there's probably more permutations than is really necessary and partly because there's a bunch of permutations that are intended to fail (if the kernel length is less than the stride length) that still have to go to all the trouble of building and saving an NN that they will never use. If this is slowing you down, feel free to take away some of these permutations or I can also open up a quick PR to address the NN issue as well

Oh wow yeah they've been taking a while. Are there specific permutations less important than others?

alecgunny:
Yeah now that I'm looking at these with fresh eyes I can definitely pare these down pretty substantially. Now that I have inference working I'll assign this as an issue to myself and get it dealt with tomorrow, should be pretty quick. Let's go ahead and merge this as-is.

alecgunny (approved):

PR #90, opened 2022-06-21
generalize sample_kernels to take trigger distance size

EthanMarx:


libs/data/bbhnet/data/utils.py
```diff
@@ -66,14 +69,31 @@ def sample_kernels(
...
     # if we're doing >1D sampling
-    min_sample_start = max(
-        x.shape[-1] // 2 + 1 - trigger_distance_size - (size / 2), 0
-    )
-    max_sample_start = min(
-        x.shape[-1] // 2 - 1 - (size / 2) + trigger_distance_size,
-        x.shape[-1] - size,
-    )
 
+    # TODO: not sure if theres a logical way to
```
alecgunny, line 73: I don't think there is, or if there is I'm not sure it would be more clear than this, so I think this makes sense. The one thing I might recommend is setting something like `dim = x.shape[-1]` to save some keystrokes and hopefully even some lines.

libs/data/bbhnet/data/utils.py
```diff
@@ -66,14 +69,31 @@ def sample_kernels(
...
+        max_sample_start = min(
+            x.shape[-1] // 2 - 1,
+            x.shape[-1] - size,
+        )
+
+    # trigger must be trigger_distance_size away from
+    # either kernel edge
+    elif trigger_distance_size <= 0:
+        min_sample_start = max(
+            x.shape[-1] // 2 + 1 - trigger_distance_size - size, 0
```
alecgunny, line 91: This should be `+ trigger_distance_size` right? Because that way the leftmost edge will always be  `< size` away from `t0`

libs/data/bbhnet/data/utils.py
alecgunny: Maybe even add in `t0 = dim // 2 + 1` to make things extra clear blow but might be overkill

libs/data/bbhnet/data/utils.py
```diff
@@ -66,14 +69,31 @@ def sample_kernels(
...
+        )
+        max_sample_start = min(
+            x.shape[-1] // 2 - 1 + trigger_distance_size,
```
alecgunny, line 94: Likewise, this should be `- trigger_distance_size` I think

libs/data/bbhnet/data/utils.py
EthanMarx: So If i remember correctly its a `+` because trigger_distance_size is negative in this instance. Does that make sense to you? If not I'll re think this.

libs/data/bbhnet/data/utils.py
EthanMarx: See above comment

libs/data/bbhnet/data/utils.py
```diff
@@ -66,14 +69,31 @@ def sample_kernels(
...
+    # combine this if else statement.
+    # Was spending too much time thinking about it
+
+    # allows for kernels that don't contain trigger
+    if trigger_distance_size > 0:
+        min_sample_start = max(
+            x.shape[-1] // 2 + 1 - trigger_distance_size - size, 0
+        )
+        max_sample_start = min(
+            x.shape[-1] // 2 - 1,
```
alecgunny, line 83: This should have a `+ trigger_distance_size` I think, so that you can sample with a leftmost edge that's to the right of `t0`

EthanMarx: Ah yes I was thinking we enforce that the kernel never goes beyond t0. Of course there is some signal (the ringdown) beyond, butI was envisioning the use case of BNS's where the majority of the signal is _before_ `t0`. If we used `trigger_distance_size` symmetrically, there would be a lot of kernels to the right of `t0` that dont contain any signal. Of course this can be alleviated by incorporating asymmetric `trigger_distance_size` 
 (e.g. `trigger_distance_size_right` and `trigger_distance_size_left`). Let me know what you think.

libs/data/tests/test_utils.py
```diff
@@ -148,9 +150,44 @@ def test_sample_kernels_with_trigger_distance(trigger_distance_size, size):
         # the trigger t0 is half way through timeseries
         t0_value = xsize // 2
 
-        # get the center of the sampled kernel
-        kernel_center_value = (np.max(kernel) - (size // 2)) % xsize
+        # get the distance of the closest sample in the kernel
+        minimum_dist = np.min(np.abs((kernel % xsize) - t0_value))
```
alecgunny, line 154: I think the logic for implementing both negative, zero, and positive `trigger_distance`s in a single test should be pretty straightforward, something like

```python
diffs = (kernel % xsize) - t0_value
close_edge = np.abs([diffs.min(), diffs.max()]).min()

if trigger_distance_size >= 0:
    assert close_edge <= trigger_distance_size
else:
    assert close_edge >= abs(trigger_distance_size)
```

Then you can just parameterize `trigger_distance_size` either as a fixture or with `@pytest.mark.parametrize("trigger_distance_size", [...])`

alecgunny (commented):
Couple implementation questions and then some aesthetic things on the testing

libs/data/bbhnet/data/utils.py
alecgunny: Ahhh of course, it's not symmetric in our case. I think your idea actually makes a lot of sense. The one catch would be that we should make sure to define the "trigger" for glitches the same way: at the _end_ of the glitch, that way this sampling method has the intended behavior for both these types of signals.

Then again, if you think marking cases that contain _just_ ringdown as `1` is valuable, then your proposed asymmetric method might make sense as well.

alecgunny (approved):

PR #104, opened 2022-06-30
First full pipeline run

EthanMarx:
Attempt at launching a full end to end run. This PR will address minor bugs that occur along the way.


projects/sandbox/export/pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = ["Alec Gunny <alec.gunny@gmail.com>"]
 
 
 [tool.poetry.scripts]
-export-model = "export:export"
+export = "export:export"
```
alecgunny, line 10: I would be careful with this, bash already reserves `export` and as I recall with DeepClean it used to complain when I originally called the script this

EthanMarx: Ah, okay. I'll change the call in the pyprojec.toml then.

EthanMarx:
closes #108

EthanMarx:
closes #121

EthanMarx:
Major components of this PR:

1. Analysis Project
2. Timeslides are now done circularly per Segment 
3. Some parallelization in the timeslide generation

@alecgunny This PR incorporates #107 #114 #107. I'm not sure if you'd rather review those separately, merge them, and then merge this, or close those and review them all in here. Let me know.

alecgunny (approved):
Looks great. Couple quick things to note as future issues but which aren't worth addressing here:
- There's an extra empty `projects/sandbox/analyze/test.py` that it doesn't seem like you need
- It's probably worth thinking about the best way to handle the prior file for the `sandbox` project. Either have both the projects that need a prior always assume relative paths to their own directories, or if we want a single unified prior, host it in `projects/sandbox` and refer to it via an environment variable (which for most use cases will be easily done at call time via `PRIOR_FILE=$PWD/nonspin_BBH.prior`, but still allows for calling the pipeline from a different directory).
- You already note in a TODO in the `analyze` project that we don't need to reload the segments, and can fit the `ClusterDistribution` for each `norm_seconds` via an `add_done_callback` on the integration `future` as soon as its ready. I would even extend this to say that we're probably so bottlenecked on data loading/writing that it's probably not worth submitting integration, fitting, or characterization jobs to the execution pool: just do them in the main process while files load in the background. The problem is that the pool will be saturated by loading jobs, so you won't even kick off any of the e.g. integration jobs until all these loads finish. You'll of course blow through them very quickly, but if you had been doing them in the main process in parallel to these load jobs it's very possible that you'll be able to keep up with the throughput and basically be done as soon as the last frame is loaded.

PR #123, opened 2022-07-21
Deterministic validation

alecgunny:
Relatively significant overhaul to support deterministic validation dataloading. At a high level, the proposed implementation looks like this:

- `WaveformSampler` and `GlitchSampler` are equipped with a `deterministic` keyword argument that dictates their behavior. If it's false, their behavior stays the same as before. If `True`, their `sample` methods are changed such that the first `N` events are selected (a value of `N=-1` selects all events) with the "trigger" placed in the center of the kernel. The `trigger_distance_size` argument has been replaced with an `offset` argument that in the `deterministic=True` case determines how far offset from the center of the kernel the triggers should be. This has the benefit that future extensions to the `DeterministicWaveformDataset` could iterate over multiple values of `offset` to get a richer sampling.
- For the `WaveformSampler`, the SNRs are normalized to all be the geometric mean of `min_snr` and `max_snr`. `generate_waveforms` now accepts a `valid_frac` argument that sets aside a certain number of the generated waveforms for validation, and saves the sampled extrinsic parameters to the saved HDF5 file. These parameters are used in place of the `priors` at dataloading time.
- The `WaveformSampler` is now initialized with `t0` and `duration` arguments which specify the `"geocent_time"` sampling range, rather than inferring these from the background. This both allows for more robust sampling of the extrinsic waveform parameters, and allows the `.fit` method to more closely reflect the API for the whitening transform (though still not as close as I would like).
- The validation `WaveformSampler` is fit to the _training_ background ASDs. I figure this allows for a better estimate of the background, since presumably the training data will be a fair bit longer than the validation. To reflect the fact that we expect these samplers to be fit on arbitrary background, `RandomWaveformDataset` will now only fit `waveform_sampler` if its `background_asds` attribute is `None`.
- The actual validation dataloader is the new `bbhnet.data.dataloader.DeterministicWaveformDataset` class. This class will sample all the waveforms and glitches up front (according to the behavior specified above) and move them to torch `Tensor`s. It will iterate through:
    - all the background data at some fixed stride (which is an input parameter)
    - all the glitches, inserting them on a one-ifo basis while iterating through the background as needed, repeating if necessary
    - all the waveforms, injecting them into the background of both ifos and iterating through the background as needed, repeating if necessary
- This can be extended in the future to accommodate time shifts in the background, coincident glitch insertion, offsets in the trigger locations, etc., but should hopefully be good enough to get us started.

Remaining work includes:
- [x] Finishing the iteration logic for `DeterministicWaveformDataset`
- [x] Adding a `validation_stride` argument to the `train` project
- [x] Adding `valid_frac` functionality to the `glitch_generation` project
- [x] Tests for `WaveformSampler` deterministic behavior
- [x] Tests for `GlitchSampler` deterministic behavior
- [x] Tests for `DeterministicWaveformDataset`


alecgunny:
Things have gotten extremely ugly here, and have made me completely rethink the way we want to do dataloading/augmentation/preprocessing. However, I figured it was more important to get this PR in so that we can start to get a rough sense of the validation curve, then overhaul the data picture in parallel. I'll open an issue outlining my proposals once this gets merged.

Both the code for iteration through the deterministic dataloader and its tests are so gross that I almost want to apologize to anyone forced to review them, but they should get the job done in the short run.

The only major change in behavior here from the proposal outlined above is that rather than create validation datasets during any of the data generation, all of the dataloader-adjacent classes now accept a `frac` kwarg which indicates which fraction of the associated objects to load in. Negative values for `frac` take samples from the back of the dataset, so the `train` script is now equipped with a `valid_frac: Optional[float]` kwarg which passes `frac=(1 - valid_frac)` to all the train dataloaders and `frac=-valid_frac` to all the validation dataloaders. This way, every run reads from the same datasets, and gets to make its own decisions about how much to actually fit the model on, and how much to use for model selection.

alecgunny:
There probably needs to be some checks to keep `frac` in range (`-1 < frac <= 1`), and same for `valid_frac` (`0 < valid_frac <= 1`), and I'm missing a test for the case that the deterministic data loader uses _both_ waveforms and glitches (which I'll admit feels silly since that's the exact case we're targeting), but the reality is I want to turn around and change everything so soon that I'm not sure it's worth continuing to get mired down in the details of this frankly bad implementation.

EthanMarx (approved):

PR #130, opened 2022-07-25
Coinc glitch

rraikman:


libs/data/bbhnet/data/glitch_sampler.py
```diff
@@ -23,8 +23,17 @@ def to(self, device: str) -> None:
     def sample(
         self, N: int, size: int, trigger_distance_size: int = 0
     ) -> np.ndarray:
-        num_hanford = np.random.randint(N)
-        num_livingston = N - num_hanford
+        if N == 0:
```
EthanMarx, line 26: I _think_ this check is redundant, as it is checked [here](https://github.com/ML4GW/BBHNet/blob/main/libs/data/bbhnet/data/dataloader.py#L169). i.e., it's never even possible to call `glitch_sampler.sample` with `N = 0`.

EthanMarx: I don't think this function needed any changes for this PR. The only logic that needed adjusting is how the glitches are inserted into the background in the `__next__` function

EthanMarx: So I think you can remove a lot of those if else statements and have:

```
num_hanford = np.random.randint(N)
num_livingston = np.random.randint(N - num_hanford, N)
```

EthanMarx:
in light of #139, This may need some additional development once that is merged.

EthanMarx:
Closing this as addressed in #153

Issue #133, opened 2022-07-26
Save injection parameters in Analysis project


EthanMarx:
Save the injection parameters along with the analyzed outputs so we can analyze events as a function of their snr's, distances, masses, etc.

Issue #132, opened 2022-07-26
Add jitter to timeslide injections


EthanMarx:
Currently, the injections for every timeslide are injected at the same times in the Hanford detector. To increase variability, it would be nice to add a `jitter` to the injection times.

PR #138, opened 2022-07-27
Analysis project improvements

### LINKED ISSUE #135: Accelerate analysis project
Right now, the analysis project is suffering from two sources of bottleneck:
- Timeslides are being integrated, written, then reloaded for building a background distribution. This was necessary when we needed to bin our integrated network outputs since there would be too much to hold in memory at once. Now that we're clustering, this isn't necessary and we can just fit the `ClusterDistribution` as the integrated outputs become available.
- Integration, background fitting, and characterization are all being handled asynchronously in the same process pool used for data loading. The problem is that data loading is a much larger bottleneck than these processes, so the added throughput by parallelizing it is just about equal to running these tasks in serial in the main process. A more optimal implementation would limit the extent to which it tries to submit these tasks to the executor pool (even avoiding doing so entirely) and leave the pool just for reading/writing.

EthanMarx:
Closes #135 #133 #132


projects/sandbox/analyze/analyze.py
```diff
@@ -183,111 +196,65 @@ def build_background(
...
-                t=t,
-                y=y,
-                integrated=integrated,
-            )
-            future.add_done_callback(
-                lambda f: pbar.update(write_task_id, advance=1)
-            )
-            fname_futures[norm].append(future)
-            segment_futures.append(future)
+                write_futures.append(future)
```
alecgunny, line 248: I think you'll want to include a `wait` for these write futures at the end of this function yeah? Unless we get some speed improvement from starting to move to characterization while the last events finish writing.

projects/sandbox/analyze/analyze.py
```diff
@@ -183,111 +196,65 @@ def build_background(
...
+                # for a given normalization window
+                shift_dir = get_write_dir(
+                    write_dir, shift, "background-integrated", norm
                 )
-                integration_futures[(norm, shift)] = [future]
 
-            # advance the task keeping track of how many files
-            # we've loaded by one
-            pbar.update(load_task_id, advance=1)
+                future = thread_ex.submit(
```
alecgunny, line 235: I feel like when I've tried to use a thread pool for reading/writing in the past I haven't seen particularly good performance, but it's been a while. Have you observed that this is able to work pretty well?

projects/sandbox/analyze/analyze.py
```diff
@@ -173,6 +181,11 @@ def build_background(
             y, t = seg.load("out")
```
alecgunny, line 181: Might be worthwhile to just have `load_segments` return `y, t`, since it doesn't seem like we need the segment object for anything once the data's been loaded

projects/sandbox/analyze/analyze.py
```diff
@@ -455,110 +422,98 @@ def analyze_injections(
...
                 write_future.add_done_callback(
                     lambda f: pbar.update(write_task_id, advance=1)
                 )
 
+                # get params file for injections for this timeshift
+                param_file = data_dir / shift / "injection" / "params.h5"
+                with h5py.File(param_file) as f:
+                    # TODO: we seem to be systematically off by 1.5 seconds;
+                    # need to find discrepancy
+                    event_times = f["geocent_time"][()] + 1.5
```
alecgunny, line 459: I think the event times are right, it's the `t` array that gets written from inference that's off. `integrate` adds `kernel_length` to the `t` array because in Tri's original implementation, `t` corresponded to the time at left edge of the kernel, not the right. The current inference code accounts for this, so that should be taken out of `integrate`.

The 0.5 is more interesting: the leading edge of the kernel that gets streamed to the server will always be `fduration / 2` ahead of the leading edge of the kernel that gets passed to the NN due to the center slice in the whitening operation. In principle this means we have a baked-in `fduration / 2` second latency in production, but since the low-latency frames are written in 1s increments this shouldn't matter too much functionally. But here, we should expect the timestamp at which the network starts firing to be `fduration / 2` seconds _after_ the event time. So it might make sense to expose `fduration` as an arg here and subtract that from `t` (or we can leave `t` as-is and include the 0.5s extra latency in the characterized `latency` array)

alecgunny (approved):
Couple points of feedback but if everything's working, nothing that should hold up moving forward with this. Mostly wondering about the performance impact of writing via the thread pool.

projects/sandbox/analyze/analyze.py
EthanMarx: So over the 10 timeslides the performance is basically the same (~ 40 seconds total for end to end analysis). I'm not sure how this will scale as we start analyzing more timeslides.

projects/sandbox/analyze/analyze.py
EthanMarx: Ah, that all adds up very nicely

EthanMarx: I'm wondering if it would make more sense to take this into account in the inference code, so that we don't have to remember to perform this shift every time we do some analysis on network outputs.

EthanMarx: Also, in the current inference code, is the shift by the kernel length taken care of on the `hermes` side?

alecgunny: So for the kernel length shift, the inference code actually doesn't do anything. It loads in the `t` array and `x` arrays, downsamples `t` to the inference sampling rate (just a strided slice), and then pairs this up with the array of network responses `y` for each `x`. The key is that each `y` is the network's prediction on a kernel for which the corresponding `x` stream (and timestamp from `t`) are on the right edge of the kernel, and the "shift" really comes from different ways of defining the timestamp of a kernel.  `integrate` adds `kernel_length` _again_ because Tri's code defined `t` differently, so really the fix here is to make `integrate` simpler and more general by taking that extra addition out.

As for the `fduration / 2` shift, that poses a more interesting dilemma, because from the viewpoint of identifying detection latency, that shift is very real: our network will actually fire `fduration / 2`s after the coalescence time, and so you could argue it's worth documenting that latency. However, for reconstructing the event time (or the time of some glitch that causes an erroneous trigger), you're right that it could be obnoxious and laborious to have to account for this every time. If the latency analysis is less important because we'll just take the max of the integrated NN output over some window, then agreed it will make more sense to just account for this in the inference code.

projects/sandbox/analyze/analyze.py
alecgunny: The same as in it doesn't matter whether you use the thread or process pool for this?

projects/sandbox/analyze/analyze.py
EthanMarx: Okay yeah totally agree that keeping track of that incurred latency is super important especially once we start running online applications where 0.5 seconds makes a world of difference. For these 'offline' applications, I still think it makes more sense to save to disk the network outputs associated with the times that BBHnet actually uses to make decisions, and treat that extra `fduration / 2` as some additional processing latency.

projects/sandbox/analyze/analyze.py
EthanMarx: Yes

alecgunny: Got it ok sounds good

projects/sandbox/analyze/analyze.py
alecgunny: Yeah I think I'm on board with that, sounds good

PR #139, opened 2022-07-29
Generalize dataloader

### LINKED ISSUE #84: Use `mldatafind` for background data read/write
Finally started playing with the background generation project this weekend and noticed a couple minor things that could be a good first issue for someone to address:
- The saved timeseries has no info about the sample rate. It might make sense to start using the `bbhnet.io.h5` submodule's `write_timeseries` functionality to save this in a more standardized format. Note that the `RandomWaveformDataset` object will need to be updated to get `t0` from the first element of the returned `t` array. It should also perform some check to ensure that the read sample rate (`1 / (t[1] - t[0])`) matches the sample rate passed to initialization.
- The logs are written to the same place the data is written. I think these will, in general, be different places. E.g. running the same training pipeline twice should produce different logs written to different directories containing all the logs for a given pipeline run, but will still read (or write) data from the same place. Maybe introduce a `data_dir` kwarg to the function which controls where data gets written, or continue to use `outdir` but add a `log_file` kwarg that can be a full path

EthanMarx:
Closes #84. 

In addition, this PR generalizes the dataloader and glitch sampler to take an arbitrary number of ifos. The ifos are
inferred directly from the background data file.


EthanMarx:
Closing this. Issue still needs addressing but will look much different than this implementation given new updates.

alecgunny:
FWIW, the dataloader itself is fully agnostic to the number of interferometers present in the data array passed to it, and the glitch sampler is (I think) largely agnostic as well. The ifo dependencies at this point are mostly in the `train` project itself, which I think could be generalized pretty straightforwardly by putting all the strain data into a single file, with mappings from IFO names to timeseries, then exposing that as a single `background` argument and then having an optional `ifos: Optional[Iterable[str]] = None` argument to `train.train:main` that if `None` uses all the datasets in the `background` file in alphanumeric order, and otherwise uses just the datasets specified to `ifos` in the specified order (or maybe we still sort them to keep things consistent). Would still require some rejiggering the `train` project to use these new args correctly but shouldn't be too bad.

PR #148, opened 2022-08-24
Fixing issues in timeslide creation

alecgunny:
Notable changes:
- Segments are queried up front, and data for each segment is downloaded in parallel to the generation of the waveforms for _all_ the timeslides for that segment
- The naming structure for timeslide root directories now prepends the interferometer initial to the shift, e.g. `dt-H0.0-L1.0`
- Per-ifo shift values have been fixed, and shifted data is clipped from each timeslide to avoid discontinuities. To keep timeslides all the same length, each timeslide is clipped to a length `max_shift = max(shifts) * n_slides * sample_rate`.
- `n_slides` timeshifts are created for _each_ interferometer that has a shift value >0, with the cartesian product of these shift values used to create timeslides. So for example, if `ifos=["H1", "L1", "V1"]`, and `shifts=[0, 1, 1]`, this would generate `n_slides**2` time shifts corresponding to `H0.0-L1.0-V0.0`, `H0.0-L0.0-V1.0`, etc.
- `prior_file`, if not an absolute path, is assumed too be relative to `__file__`. `projects/sandbox/pyproject.toml` has been updated to reflect this
- A `min_segment_length` kwarg has been added to keep from creating segments that won't have enough data to be meaningful, and a value of `1024` has been added to `projects/sandbox/pyproject.toml`
- A `test_stop` argument has been added to the `[tool.typeo.base]` table in `projects/sandbox/pyproject.toml`. Now the timeslide generation will happen on a true test set, from `base.stop` to `base.test_stop` (which amounts to ~90k seconds worth of segments).

The major bottleneck here is now the waveform projection, and it's making things pretty slow. But since we have a fast implementation of that coming down the pipe in the next couple days, I'm willing to let that coast for now since you only in principle need to run this script once anyway.


projects/sandbox/timeslide_injections/timeslide_injections.py
```diff
@@ -108,110 +191,196 @@ def main(
...
+            )
 
-        # create TimeSlide object for injection
-        # this will create the directories if
-        # they don't exist
-        injection_ts = TimeSlide.create(root=root, field="injection")
+            # generate timestamps of trigger times for each waveform.
+            # Jitter will get added in `get_params`
+            signal_start = segment_start + buffer_
+            signal_stop = segment_stop + buffer_
```
EthanMarx, line 260: Should be `signal_stop = segment_stop - buffer_`

EthanMarx (approved):
Looks good. One minor fix on generating the signal times w/ the `buffer`. 

One thing we should keep in mind moving forward is that by cutting off `max_shift * n_slides` for each segment, we limit our segment sizes, especially once we start doing ~ 1000s of slides. I guess this is the price we pay for doing circular segments.

projects/sandbox/timeslide_injections/timeslide_injections.py
alecgunny: Good catch, fixed

alecgunny:
Yeah this is a good point. Probably not something we need to worry about for the immediate future because we can always just extend our search window to avoid having to do too many slides, but for O(1000 months) background this will obviously start to become impractical.

This is also why I don't compensate for this when comparing against `min_segment_length`, the assumption for the time being that either the user can account for this when setting `min_segment_length` or that even if they don't the difference will likely be small compared to the desired minimum segment length, but that would need to be fixed in the future as well.

Issue #111, opened 2022-07-01
Rewrite injection library


EthanMarx:
I have started putting together some of the code for the [PE](https://github.com/ML4GW/PE) project, and naturally have seen opportunities for generalization, specifically with the `injection` library. This might be a good time to consider how we may implement this in its own repo. Is the injection code large enough to warrant its own repo? If not, what else should go in? related to ML4GW/PE#6

wbenoit26:
I feel that it's large enough on its own at this point, but there are also things that could be added. At the moment, it feels pretty rigid in how it gets used. Generalizing to broader use cases, and maybe making it a bit more user-friendly, could be good improvements to add.

EthanMarx:
I totally agree. 'm a bit confused about the logic behind some of the recent changes we've made to the library. For example, I  liked how the old `generate_gw` expected a fully initialized `WaveformGenerator`. It gets a bit confusing now with the `**waveform_generator_args`, as its not exactly obvious what arguments these can be. Additionally, many of the defaults are hidden behind the `get_waveform_generator` function. 

I think the idea behind these changes was to make [these](https://github.com/ML4GW/BBHNet/blob/main/projects/sandbox/generate_waveforms/generate_waveforms.py#L67#L69) arguments in `generate_waveforms` more general. I think simply passing them to the `generate_waveforms` main function, and then explicity initializing the `waveform_arguments` dict in the `WaveformGenerator` makes the most sense. That way, `generate_gw` is as most general as possible, since the details of the waveform generator are left to the user.

@deepchatterjeeligo let me know your thoughts.

deepchatterjeeligo:
Overall I agree that a redesign is needed. There are several function which are rather long and have parts that are hard to test. But also I have been looking at this code rather recently. But to answer

> It gets a bit confusing now with the **waveform_generator_args, as its not exactly obvious what arguments these can be.

It is all the keyword arguments that go into the bilby waveform generator. In our code, we are wrapping around a bilby function and hence do not need to add them once again to the function signature. If in a new release of bilby, the keyword argument changes to something else, we should don't need a change in our end. We can assume that the user will take note that the bilby version has changed. So I think the docstring should contain a cross ref to the corresponding bilby docs to know what the arguments can be.

However, it is true that the code tries to unwrap a nested dictionary `waveform_arguments`. While I made that change, I don't feel very strongly about it since it break the logic I mentioned above. But I thought that was a preference in #70

[...]

PR #150, opened 2022-08-30
Injection rewrite

### LINKED ISSUE #122: Expose waveform generator attributes as raw arguments in `injection` functions
A couple of the functions in the `bbhnet.injection` expose a `waveform_generator` argument solely for the purposes of extracting attributes from it like `sampling_rate`. I'm thinking specifically of [these lines](https://github.com/ML4GW/BBHNet/blob/820edd361bae527e445da1308a62e50293bec8d6/libs/injection/bbhnet/injection/injection.py#L93-L94) in `project_raw_gw`. It might make more sense to expose these arguments explicitly, and let the calling code grab these attributes from a waveform generator if it wants.

`bbhnet.injection.utils.apply_high_pass_filter` _would_ have the same drawback, except that it has the actual waveform generation functionality wrapped up inside of it. Would it make sense to split the waveform generation off into its own function that generates a `(num_waveforms, num_ifos, waveform_size)` sized array, then a separate function for doing the high pass filtering along this array's `axis=-1`? In this case, the filtering function should similarly expose the `sampling_frequency` and `minimum_frequency` arguments explicitly.

EthanMarx:
Rewrite of injection library. Most functionality has been removed as it is being implemented [here](github.com/ML4GW/ml4gw). 

- [x] Rewrite `timeslide_injections.py` to use new `generate_gw` function
- [x] Use ml4gw `project_raw_gw` in `timeslide_injections.py`
- [x] rewrite `generate_waveforms.py` to use new `generate_gw` function
- [x] Add torch accelerated snr calculation to `timeslide_injections.py` once ML4GW/ml4gw#1 is merged
- [x] Incorporate ml4gw WaveformSampler 
- [x] `apply_highpass_filter` has been removed. Use `torchaudio` module (or some torch implementation) instead. Maybe bake this into the model somehow (like the `WhiteningTransform`) to ensure all data is being treated the same?
- [x] Write tests for array like injection 

closes #122 #111


EthanMarx:
@alecgunny @wbenoit26 I think the inconsistency Alec was referring to between `generate_waveforms` and `timeslide_injections` is resolved here.

EthanMarx:
closes #157 and #160

alecgunny (approved):

Issue #128, opened 2022-07-25
Allow Distributions ability to incorporate vetoes


EthanMarx:
Allow the `Distribution` class the ability to take a list of veto segments and disregard background events that lie in these segments. I'm not sure what the best way to implement this is. For example, the `Cluster` distribution class assumes contiguous times, so removing times would, at the moment, be problematic. One basic approach would be to set the network outputs for all vetoed times to `-inf` in the `Distributions` `.fit` method.

alecgunny:
Would it be sufficient to handle this in the parent class's `.fit` method by breaking a `Segment` up if necessary into multiple sub-segments, then running `.update` on each one of these new, shorter segments?

Issue #149, opened 2022-08-25
Distribution class rewrite


EthanMarx:
Been thinking about the Distribution classes. Here are some ideas for restructuring:

1. Retire the `Discrete` class: I think ultimately it is important to be able to identify  background ‘events’  associated with specific times in each detector. Only storing binned events loses this information.
2. I think we can keep the base `Distribution` class for now. The `update(y, t)` will become `update(y, t shifts) `to be able to track event times. 
3. Add an `apply_vetoes` method that removes event times from the distribution within vetoed segments
4.  Make the `characterize_events` method a function separate from the Distribution  class.  How exactly we analyze events seems disparate from the background distribution used to analyze them. I still think the FAR vs latency curves the current `characterize_events` method uses is a useful metric. However, for recovering events in say, some blind challenge where we don’t know the event times, we’ll have to perform the same event time identification method (e.g., clustering) we are using for the background to identify ‘events’. 
5. Since we know all of the injection times for our current timeslide/injection analysis, we are able to choose the optimal event time (i.e. once the trigger leaves the kernel) to use for the ROC curve. However, we won't know this information in real time (ideally, our clustering algorithm or whatever event identification we use should be able to do this optimally, but this needs to be tested!). We should be treating our background analysis and injection / 0-lag analysis on equal footing. 
6. The current `Cluster` class performs clustering by making consecutive windows, and taking the maximum within each window. This is not necessarily optimal, as the window may ‘split’ a glitch (or a GW signal), hence double counting it. I talked with Cody Messick from the gstlal group and here is how they perform clustering: 

    “For clustering, we take the max likelihood-ratio (our ranking stat) in any given 8 second window, i.e. the result of our clustering is that there is no event closer in time to any other event by 4 seconds, and each remaining event is the highest LR (hence lowest FAR) in an 8 second window centered on itself”
    
Open to any thoughts or suggestions

PR #151, opened 2022-08-30
Rewrite distribution

EthanMarx:
- [x] Retire `characterize_events`
- [x] Retire `Discrete` Class
- [x] Write clustering algorithm for `Cluster` class
- [x] Add `apply_vetoes` method to `base` class that removes event times in vetoed segments
- [x] Write tests for apply_vetoes method
- [x] Write tests for clustering algorithm


EthanMarx:
1. The `Segment.shift` and `TimeSlide.shift` properties are replaced by `.shift_dir`, which will return something like `dt-H0.0-L4.0`. `Segments` and `TimeSlides` now have new `.ifos` and `.shifts` properties, which will parse the .shift_dir for the `ifos` and `shifts`. For example, A TimeSlide with a `shift_dir` value of `dt-H0.0-L4.0` will return a`.ifos` value of ['H', 'L'] and a .shifts value of [0.0, 4.0]. With these new additions, we should consider enforcing this directory naming convention when initiating new `TimeSlides`, or appending `Segments`
2. The `characterize_events` method and `DiscreteDistribution` have been retired
3. The base `Distribution` class initializer now also requires the `ifos` used to generate the events. The `update` method of the base `Distribution` class now requires a `shifts` argument to ensure we can keep track of event times properly. The order the shifts are passed corresponds to the order the `ifos` are passed in the initalizer.
4. An `apply_vetoes` method has been added to the base `Distribution` class. This function takes arguments `**vetoes`, where the argument name is the `ifo` and the value is a list of vetoed segments. From the `event_times` and the `shifts`, the function will apply the corresponding time shifts for the given interferometer, and remove any events that lie in the vetoed segments.

EthanMarx:
closes #128 #149

alecgunny (approved):

PR #153, opened 2022-09-07
Training acceleration and improvements

### LINKED ISSUE #72: Two-glitch insertion for RandomWaveformDataset
`bbhnet.data.RandomWaveformDataset` only supports inserting glitches in one ifo channel at a time, which mitigates the network's ability to ignore coincident glitches. One way to fix this might be to sample like:

```python
num_hanford_glitches = np.random.randint(self.num_glitches)

# sample at a minimum enough livingston glitches to achieve
# at least self.num_glitches, but at most self.num_glitches
num_livingston_glitches = np.random.randint(self.num_glitches - num_hanford_glitches, self.num_glitches)

# ... do sampling
hanford_glitches = ...
livingston_glitches = ...

background[:num_hanford_glitches, 0] = hanford_glitches

# coincident glitches are inserted over indices
# [self.num_glitches - num_livingston_glitches, num_hanford_glitches]
# so if num_livingston_glitches = self.num_glitches - num_hanford_glitches,
# there will be no coincident glitches in the batch
background[self.num_glitches - num_livingston_glitches: self.num_glitches, 1] = livingston_glitches
```

### LINKED ISSUE #136: Implement `WaveformSampler` and `GlitchSampler` as torch `Module`s
One way to implement these data-augmenting objects, rather than to incorporate them directly into the dataloader, would be to implement them as torch `Module`s that get used as part of a preprocessing `Module` and use the `self.training` attribute to decide whether to apply their transformation or just return the tensor as-is.

This would let the dataloader behave much more simply and make it easier to consolidate the behavior validation and training dataloaders. It would also allow these modules to be much more composable and used by other project.

### LINKED ISSUE #137: Rethink data loading and validation
Especially in the context of #136 , it might make sense to start stripping back some of the omnibus functionality of the dataloaders into pretty simple array-iteration objects, with the option to `shuffle` that iteration. Importantly, rather than loading in the data itself, the data would get loaded in e.g. the `train` project then passed as arrays to the data loader.

The validation dataloader could then look just like the training dataloader, except that the array it used would be a concatenation of the background data pre-batched with copies of that data with the modules outlined in #136 applied to them first (and of course it wouldn't shuffle how it iterated through this data). The dataloader could include logic in the `__init__` about whether the dimension of the data it was passed was 2 or 3 (3 indicating pre-batched data). This will also have the benefit of providing a smoother transition to out-of-memory dataloaders should we ever need to build them.

### LINKED ISSUE #143: Better waveform SNR sampling
As part of #136 , rather than uniformly sampling waveform SNRs between `min_snr` and `max_snr`, accept an `snr_prior` argument which is some sort of distribution object (either from bilby or scipy) which gives more flexibility to sample lower-SNR events

alecgunny:
Notable changes:
- Using `ml4gw.transforms.RandomWaveformInjection` transform for computing injections during data loading.
- Reformatted `GlitchSampler` to follow similar structure as `RandomWaveformInjection` to combine into single `augmenter` module which gets called inside dataloader
- `GlitchSampler` and `RandomWaveform` injection have no information about what the other is doing, therefore glitches can be inserted into samples with waveforms. Moreover, the new `GlitchSampler` inserts glitches into IFOs independently, meaning there can be coincident glitch insertion as well.
- Using light subclass of `ml4gw.dataloading.InMemorydataset` for doing dataloading, which adds a `y` tensor of all zeros and calls an optional `preprocessor` module on both `X` and `y`. The `augmenter` module gets used in this.
- Adding `bbhnet.data.distributions` submodule for sampling source parameters. For `snr`, using a log normal distribution whose parameters are supplied to `train.main` (mean and standard deviation of the desired distribution, with an optional min for clipping values that go too low. I wanted to use the mode rather than the mean, but the math doesn't work out as nicely).
- Adding a highpass filter module to the preprocessor. This is currently bottlenecking the pipeline.

TODOs:
- [x] Implement validation by sampling validation waveforms and glitches up front and inserting them into a validation background. Use a simple tensor dataset to iterate through the pre-computed and concatenated background, waveforms, and glitches.
- [x] Highpass filter doesn't currently do any windowing and probably has some other bad practices as well that might be improved by adopting practices from scipy or gwpy
- [x] Highpass filter is currently bottlenecking training due to the fundamentally recursive nature of the IIR component of the response. Once validation is implemented, worth doing a test to confirm this is actually helping things, as it's slowing training down by almost a factor of 2.
- [x] Tests for `bbhnet.data.distributions`
- [x] Test for highpass filter
- [x] Merge `ml4gw` branch this is currently using (REPLACE WITH PR LINK WHEN IT'S MADE)
- [x] If we're keeping the highpass filter, implement it as part of preprocessor at export time
- [x] Document new arguments to `train.main`
- [x] Document `bbhnet.data.distributions`
- [x] Document highpass filter
- [x] Remove existing `WaveformSampler` implementation

Closes #143 , closes #136 , closes #137 , closes #72


projects/sandbox/train/train.py
```diff
@@ -1,17 +1,207 @@
...
+    # project them to inteferometer responses and
+    # inject those resposnes into the input data
+    with h5py.File(waveform_dataset, "r") as f:
+        signals = f["signals"][:]
+
+    # TODO: right now signals are prepared such that the
+    # coalescence time is at the end of the window, so roll
+    # them to put them in the middle as expected. Do we want
+    # to do this here or in the generate_waveforms project?
+    signals = np.roll(signals, -signals.shape[-1] // 2, axis=-1)
```
EthanMarx, line 144: Once #150 is merged, this will be done in `bbhnet.injections`

alecgunny:
Once this is merged, open up an issue discussing the high-pass filter problems encountered here and suggest avenues of investigation

alecgunny:
New modularized training project needs tests for individual functions, but this seems to be working so if this is holding up other PRs I suggest we merge without it and open up an issue for it as a good first issue

libs/data/bbhnet/data/glitch_sampler.py
```diff
@@ -1,90 +1,45 @@
...
-                )
-                livingston = torch.stack(livingston, axis=0)
-
-        return hanford, livingston
+            )
+
+        masks = torch.rand(size=(len(self.glitches), len(X))) < self.prob
+        # masks = masks.to(X.device)
+        for i, ifo in enumerate(self.glitches):
+            mask = masks[i]
```
EthanMarx, line 34: So this mask determines which background elements are replaced with glitches

libs/data/bbhnet/data/glitch_sampler.py
```diff
@@ -1,90 +1,45 @@
...
+            N = mask.sum().item()
+            idx = torch.randint(len(ifo), size=(N,))
```
EthanMarx, line 36: And this determines which glitches get sampled

projects/sandbox/timeslide_injections/timeslide_injections.py
```diff
@@ -105,6 +105,16 @@ def inject_waveforms(
...
+        # alternatively:
+        idx = np.arange(signals.shape[-1])[None]
+        idx = np.repeat(idx, len(signals), axis=0)
+        time_diffs = signal_times - times[0]
+        idx_diffs = (time_diffs / sample_rate).astype("int64")
+        idx += idx_diffs[:, None]
+        idx = idx.reshape(-1)
+        signals = signals.reshape(-1)
+        x[idx] += signals
+
```
EthanMarx, line 117: Also implemented in #150

projects/sandbox/train/train/utils.py
```diff
@@ -0,0 +1,135 @@
...
+    to support multiple input
+    """
+
+    def forward(self, *inputs):
+        for module in self._modules.values():
+            if isinstance(inputs, tuple):
+                inputs = module(*inputs)
+            else:
+                inputs = module(inputs)
+        return inputs
```
EthanMarx, line 29: Is this to support augmenting the background with glitches and signals, as well as changing target tensors

EthanMarx (approved):
Left some questions / points of clarification. There are a couple of items that are addressed in #150. I think we merge this anyways since it works, and then I'll undo those changes in that PR.

libs/data/bbhnet/data/dataloader.py
```diff
@@ -1,476 +1,40 @@
...
-
-        # load in the background data
-        self.hanford_background = _load_background(hanford_background, frac)
-        self.livingston_background = _load_background(
-            livingston_background, frac
+        super().__init__(
+            X,
+            kernel_size,
+            batch_size=batch_size,
+            stride=1,
```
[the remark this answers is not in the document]
alecgunny: Ah ok, I think my mistake was actually probably in including this in the arguments to the dataloader, since my assumption was that we'll always want a stride of 1 for this dataloader, but open to thoughts here: correct this to be `stride=stride`, or remove the `stride` argument altogether from the `BBHNetInMemoryDataloader`?

libs/data/bbhnet/data/glitch_sampler.py
alecgunny: yeah exactly. Marginally faster to just generate one mask up front (I think, maybe the slice ends up costing you but I'm fairly confident this is the faster way of doing it)

libs/data/bbhnet/data/glitch_sampler.py
alecgunny: Exactly

projects/sandbox/train/train/utils.py
alecgunny: Yeah exactly. Part of me thinks it might make more sense to make an `Augmenter` module the same way we have a `Preprocessor` module and handle this more explicitly there (i.e. have the injector and inserter as attribute submodules of the `Augmenter` submodule and have the forward method look like
```python
def forward(self, X, y):
    X = self.inserter(X)
    X, y = self.injector(X, y)
    return X, y
```
but there's just a couple implementation details about how/where it would get initialized that I don't really want to bog this PR down with.

libs/data/bbhnet/data/dataloader.py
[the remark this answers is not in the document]
EthanMarx: Remind me: a `stride` of 1 means the dataloader will move by 1 `sample` when creating `kernels` of background data? If so, I agree. We'll probably only ever want strides of 1.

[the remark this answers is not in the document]
alecgunny: Yeah exactly, or in this case that it's capable of sampling kernels at a resolution of 1 sample. But I agree, I'll take the arg out of the dataloader then.

projects/sandbox/timeslide_injections/timeslide_injections.py
alecgunny: Oh shoot I'm fairly confident I didn't mean to implement this here, I think it's just where I mocked this idea up and forgot to take it out. I'll checkout this file back to `main`, and we'll use the implementation in #150

PR #159, opened 2022-09-27
Improving ci

### LINKED ISSUE #86: Parallelize CI
The [GW-IaaS CI](https://github.com/fastmachinelearning/gw-iaas/blob/main/.github/workflows/ci.yaml) handles path filtering a bit differently: it sets all the filters for a given library in the `filter` job, rather than creating a single filter for each library then using OR statements in the individual `if` clauses.

The benefit of adopting this would be that we could potentially adopt the matrix syntax used by the [DeepClean CI](https://github.com/ML4GW/DeepClean/blob/main/.github/workflows/ci-tests.yaml), while still getting the benefits of filtering, by referencing `{{ matrix.lib }}` in the `if` statement. This will allow our tests to run in parallel, speeding up CI.

The on tricky thing will be the ability to use per-job marks in pytesting (e.g. the `data` tests use the `-m "not gpu"` mark to ensure that we don't try to run GPU dataloader tests). There might be a way to control this with some sort of per-job environment variable, but I'm not sure what sort of support GitHub has for this.

### LINKED ISSUE #145: Don't run CI workflow on draft pull requests
Is it worth setting up our workflows so that they don't run on draft PRs, that way we aren't wasting cycles on code we know won't work? Should be as simple as adding `if: github.event.pull_request.draft == false` to the `ci-tests` job in `.github/workflows/ci-tests.yaml`, but just wanted to open up the discussion to confirm this would be desirable first?

alecgunny:
Creating a draft pull request intended for someone to be able to close #86 and close #145 in one fell swoop. The idea would be
- [x] For each library and project, create a filter indicating the libraries on which it depends (including `ml4gw` and `hermes`), and set the output of these filters as the output of the first job (see how [hermes does this](https://github.com/ML4GW/hermes/blob/94064d0920ee690432fbfbe12fb34b0b67636012/.github/workflows/unit-tests.yaml#L18-L25))
- [x] Use this output from this job to set up a matrix of libraries/projects that need to get tested in parallel in the following job (once again, [hermes](https://github.com/ML4GW/hermes/blob/94064d0920ee690432fbfbe12fb34b0b67636012/.github/workflows/unit-tests.yaml#L35))
- [x] For each library/project test job, set up any custom test arguments using the `include` syntax outlined [here](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-expanding-configurations) (e.g. `-m "not cuda"` for the dataloader tests.
- [x] In the first job, include an `if: github.event.pull_request.draft == false` check and something to check if the commit message includes `[skip-ci]` to skip the jobs altogether. Importantly, the first case should be marked as a `skip` (the idea here is that you don't want to get a bunch of obnoxious notifications about a draft PR that you don't expect to pass failing, and we can merge a draft request anyway so there's no risk), while the latter should be marked as a `fail` (this way we don't manage to get around meeting CI requirements just because we told GitHub to skip the tests).
- [x] All tests should run if any changes are made to the workflow file, to ensure that we're still running things correctly.


EthanMarx:
This is more or less done. 

Need the infer tests to be fixed up. it looks like there were some `hermes` changes that affected the mocks. Will probably be easiest if @alecgunny addresses this.

alecgunny:
Ok yeah it looks like I can set the requirement that the checks pass on the branch protection rules, so we should be all set. I just can't do it now because the name of the workflow is different from the one in `main`, so GitHub won't be able to check for it. I think `unit-tests` is a better name though, so I'm happy to stick with it and make that change once this is merged.

I'll file a separate PR to fix the issues with the `infer` tests, and then we can rebase this on top of that and merge it.

EthanMarx (approved):

PR #164, opened 2022-10-01
Generate background tests

EthanMarx:


projects/sandbox/generate_background/generate_background.py
```diff
@@ -63,9 +63,11 @@ def main(
         stop,
     )
 
+    print(segments)
```
alecgunny, line 66: Were these handful of print statements just left over from debugging? Wanted to flag them just in case

alecgunny (approved):
Wanted to note the couple of print statements in the project script but other than that looks good

Issue #155, opened 2022-09-21
improved performance metrics


EthanMarx:
The sensitive distance from the ML GW search [paper](https://dcc.ligo.org/DocDB/0183/P2200233/002/MLGWSC_1.pdf) is a good place to start. See Section C. If we adopt their injection distribution, we can compare directly to those algorithms, which includes pyCBC. They use 1 month of background which is very doable.

Longer term, we should move towards producing something like Figure 6 in the [catalog paper](https://arxiv.org/pdf/2111.03606.pdf). I need to sit down with someone to go over the details of that calculation. 

@wbenoit26

EthanMarx:
Closing as I think we now have a solid handle on this

PR #165, opened 2022-10-03
Restructuring analysis project

### LINKED ISSUE #120: Possible bug in analysis project
The `GaussianNormalizer` [here](https://github.com/ML4GW/BBHNet/blob/main/projects/o2-timeslide-analysis/analyze_o2.py#L414) is initialized with `norm` while [here](https://github.com/ML4GW/BBHNet/blob/main/projects/o2-timeslide-analysis/analyze_o2.py#L221) it is initalized with `norm * sample_rate`

alecgunny:
Closes #120 , related to #155 

- Fitting background and foreground distributions on matching segments at the same time so that normalizers only need to be fit once
- Reorganizing and restructuring project to be more clear and modular
- Visualization app issues
    - [x] Event inspector is not rendering changes to data sources
    - [ ] Glitch plot legend isn't showing up when using `legend_group` instead of `legend_field` as the argument to the circle plot. The issue with using `legend_field` as noted [here](https://docs.bokeh.org/en/latest/docs/user_guide/annotations.html#hiding-legend-items) is that Bokeh treats all the legend items as a single item, so you can't click to hide them individually - they're either all on or all off.
    - [x] x/y range limits for the efficiency vs. FAR plot aren't updating when we switch the x-axis to SNR
- Visualization app additions
    - [x] How to pass arguments? Do something like [this](https://github.com/bokeh/bokeh/blob/2.4.3/examples/howto/server_embed/standalone_embed.py) but inside of a `typeo`-fied `main` function?
    - [ ] Toggle in row with norm select to turn vetoes on/off that affects all plots
    - [x] Toggle in row with FAR/SNR select widget that affects whether efficiencies are computed for <= the range value, or just within the corresponding bin (should be reflected in the legend labels as well).
    - [ ] Additional event inspection window for Q-scan or spectrogram or something
    - [ ] Additional event inspection window for saliency map of gradient wrt. spectrogram
- Library unit tests
    - [x] Cluster distribution load `Tb`
    - [x] Cluster distribution `from_file` with `ifos`
    - [x] `integrate` w/o `kernel_length` argument
    - [x] `inject_waveform` with `sample_rate != 1` and with `idx` relative to center of waveform


alecgunny:
@EthanMarx feel free to try to tackle some of the vizapp TODOs if you want to get familiar with this framework. To run the server, connect to the LDG cluster via SSH with port forwarding enabled (`ssh -L 5006:localhost:5006...`) and then cd to the `vizapp` directory. Then just `poetry install` and `poetry run bokeh serve vizapp`, then the app should be available at `localhost:5006` on your local machine.

EthanMarx:
Event inspector has been fixed, although the plot needs some formatting / style changes.

Once the tests have been written, let's merge this in and then we can open the remaining tasks as issues.

alecgunny:
@EthanMarx give the vizapp another look before merge, made a few changes:
- Organized things into tabs, with the first tab for summary plots (right now just the perf summary, but leaving room for more if we want to add them later) and the second for more detailed analysis (background, event inspector, etc.)
- Moved some legends outside plots to avoid obscuring data. For the distribution plot, removed the legend entirely and instead rely on y-axis labels/coloring to distinguish the meaning of the glyphs.
- Changed the coloring on the distribution plot so that it's isn't confused with the green/orange used to differentiate the interferometers in other plots.
- Updated the title on the distribution plot so that it can contain all the relevant info
- Made the box select tool the default click-and-drag tool on the distribution plot (instead of box zoom)

EthanMarx:
Let me know when glitch insertion is fixed and I'll give this a look over.

alecgunny:
@EthanMarx should be good to go

EthanMarx (approved):

PR #171, opened 2022-10-17
Addressed minor bug fixes

### LINKED ISSUE #162: Miscellaneous Bugs in the Pipeline up through Timeslide Injections
1. In trainer.py, an error is raised [here](https://github.com/ML4GW/BBHNet/blob/000f666c133bfc595d8274f5bd14852feb57918e/libs/trainer/bbhnet/trainer/trainer.py#L82) because the target shape (4096) is different than the prediction shape (4096, 1). Fix this by changing [this](https://github.com/ML4GW/BBHNet/blob/000f666c133bfc595d8274f5bd14852feb57918e/projects/sandbox/train/train/validation.py#L91) line to be `y = torch.zeros((len(X), 1)))`
 
 2. The pyproject.toml in sandbox references [base.fmin](https://github.com/ML4GW/BBHNet/blob/000f666c133bfc595d8274f5bd14852feb57918e/projects/sandbox/pyproject.toml#L97). Should be `base.minimum_frequency`
 
 3. We might want to make [`clean`](https://github.com/wbenoit26/BBHNet/blob/000f666c133bfc595d8274f5bd14852feb57918e/projects/sandbox/pyproject.toml#L144) be `false` by default, just so it's easy to run this pipeline out of the box.
 
 ~~4. It looks like `max_segment_length` is no longer an argument in timeslide_injections.py, so it should be removed from the [pyproject.toml](https://github.com/wbenoit26/BBHNet/blob/000f666c133bfc595d8274f5bd14852feb57918e/projects/sandbox/pyproject.toml#L163)~~
4. `max_segment_length` in [pyproject.toml](https://github.com/wbenoit26/BBHNet/blob/000f666c133bfc595d8274f5bd14852feb57918e/projects/sandbox/pyproject.toml#L163) isn't implemented in `timeslide_injections` yet, so it should be commented out for the time being.

wbenoit26:
Closes #162


wbenoit26:
@alecgunny Is this all you were thinking for the `clean` argument?

alecgunny:
@wbenoit26 yeah exactly

alecgunny:
Now that I'm thinking about it, it might be worthwhile to have @rafia17 try to run the pipeline end-to-end from your branch once all your changes are in, just to confirm that this covers everything a new user would need to get up and running. To that end, it might make sense to change the export platform in `projects/sandbox/pyproject.toml` to `onnxruntime_onnx` to avoid the CUDA version issues that come with TensorRT for now. I can prioritize that fix in `pinto` and then we can confirm it fixes things in a late PR

projects/sandbox/generate_background/generate_background.py
```diff
@@ -53,7 +53,6 @@ def main(
             "Background data already exists"
             " and forced generation is off. Not generating background"
         )
-        return
```
EthanMarx, line 56: What was the problem here? Don't think we wan't to remove these.

wbenoit26: Yeah, I thought it might be the most straightforward fix, but now I'm less sure of that

wbenoit26:
Closes #174

projects/sandbox/generate_glitches/generate_glitches.py
```diff
@@ -254,12 +254,11 @@ def omicron_main_wrapper(
     omicron_main(omicron_args)
 
 
-@typeo
+@scriptify
 def main(
     snr_thresh: float,
     start: int,
     stop: int,
-    test_stop: int,
```
EthanMarx, line 262: Was there a reason for this? This was included to allow us to run omicron on the testing set.

wbenoit26: Oh, I see. Yeah, I don't know exactly what I've done, but I've screwed up my version of the code somehow. My version had the `test_stop` argument in the function definitions, but not in the `generate_glitches` section of the `pyproject.toml`, and not anywhere in the rest of the `generate_glitches` script, so I figured it wasn't supposed to be there.

I'll add it back in.

EthanMarx: Definitely my fault. I likely didn't add it in the `generate_glitches` config

alecgunny (approved):
Only issue is that `projects/sandbox/analyze/analyze.py` has been added back in, but it's not worth holding everything up for. Let's merge this and then turn around and delete that in another PR.

PR #180, opened 2022-10-26
Fixing issues with futures in analyze script

### LINKED ISSUE #179: `wait` call in `analyze` project is failing
Basically we're trying to append to the `write_futures` list inside of the [callback](https://github.com/ML4GW/BBHNet/blob/52ea14b50cba61dc141af232562cc7b68127173f/projects/sandbox/analyze/analyze/utils.py#L127) used for the future that [integrates the network response](https://github.com/ML4GW/BBHNet/blob/52ea14b50cba61dc141af232562cc7b68127173f/projects/sandbox/analyze/analyze/main.py#L95).

This causes a problem with `concurrent.futures.wait`, which does not expect the size of the list you pass it to change. There are a few things to note there:
- We don't really need to be using futures for the integration: what's bottlenecking things is the distribution fitting, which has to iterate through the timeseries of outputs. The process pool should be used for parallelizing this, with the integration done in the main process. However, since processes can't update objects in place, they'll each create their own distribution, which means we'll have to aggregate their collected attributes after the fact.
- We might not even need to parallelize the writing given how much of a bottleneck the fitting is, which would mitigate this issue basically entirely.
- Even if we do, we can submit the write job from the main process as soon as the integration is done and add the future to a giant future list (of `fit`s and `write`s). We can then `wait` on this, aggregate the results from all the various distributions, then write these distributions out.

alecgunny:
Closes #179


alecgunny:
Adding in vizapp fixes for #181 , adding logging to vizapp to know when app is ready to go and adding fixes for when SNR bins are empty, getting rid of logging in `ClusterDistribution` class to avoid thousands of individual log lines when running in debug. A downstream function calling `ClusterDistribution.nb` can add this log if they so choose.

EthanMarx (approved):

PR #182, opened 2022-10-26
Prior submodule

wbenoit26:
I think there's a cleaner way to implement this, but I'm not sure how. Open to suggestions.

### DIFF
--- a/libs/injection/bbhnet/injection/priors.py
+++ b/libs/injection/bbhnet/injection/priors.py
@@ -1,0 +1,98 @@
+import numpy as np
+from bilby.core.prior import (
+    ConditionalPowerLaw,
+    ConditionalPriorDict,
+    Constraint,
+    Cosine,
+    PowerLaw,
+    PriorDict,
+    Sine,
+    Uniform,
+)
+from bilby.gw.prior import UniformComovingVolume, UniformSourceFrame
+
+
+def prior_selector(prior_name):
+
+    if prior_name == "nonspin_bbh":
+        return nonspin_bbh()
+    elif prior_name == "end_o3_ratesandpops":
+        return end_o3_ratesandpops()
+    else:
+        raise ValueError(f"Prior name {prior_name} is not an option")
+
+
+def nonspin_bbh():
+
+    prior_dict = PriorDict()
+    prior_dict["mass_1"] = Uniform(
+        name="mass_1", minimum=5, maximum=100, unit=r"$M_{\odot}$"
+    )
+    prior_dict["mass_2"] = Uniform(
+        name="mass_2", minimum=5, maximum=100, unit=r"$M_{\odot}$"
+    )
+    prior_dict["mass_ratio"] = Constraint(
+        name="mass_ratio", minimum=0.2, maximum=5.0
+    )
+    prior_dict["luminosity_distance"] = UniformSourceFrame(
+        name="luminosity_distance", minimum=100, maximum=3000, unit="Mpc"
+    )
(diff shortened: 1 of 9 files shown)


libs/injection/bbhnet/injection/priors.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def prior_selector(prior_name):
+
+    if prior_name == "nonspin_bbh":
+        return nonspin_bbh()
+    elif prior_name == "end_o3_ratesandpops":
+        return end_o3_ratesandpops()
+    else:
+        raise ValueError(f"Prior name {prior_name} is not an option")
```
EthanMarx, line 22: I don't think you need this prior_selector. When we need to specify a Prior in a config, we can annotate that in the function args as a `Callable`, and in our `typeo` config pass the python path, e.g `"bbhnet.injection.priors.nonspin_bbh"`

projects/sandbox/generate_waveforms/generate_waveforms.py
```diff
@@ -1,18 +1,17 @@
...
 
-from bbhnet.injection import generate_gw
+from bbhnet.injection import generate_gw, prior_selector
 from bbhnet.logging import configure_logging
 
 
 @scriptify
 def main(
-    prior_file: Path,
+    prior_name: str,
```
EthanMarx, line 14: See above comment. Think this can just be 
```
def main(
    prior: Callable,
):
    prior = prior()
```

libs/injection/bbhnet/injection/priors.py
alecgunny: I think this is exactly right. I think there's maybe even a case for having this hardcoded into a project, e.g. `from bbhnet.injection.priors import nonspin_bbh`, but if we think this is something we'll want to have some flexibility with at the project level this works.

alecgunny: The other alternative is to just construct these dicts at the outermost level of `priors.py` and make them available via `priors` dict of dicts, i.e.:
```python
nonspin_bbh = ...
end_o3_ratesandprops = ...
priors = {
    "nonspin_bbh": nonspin_bbh,
    "end_o3_ratesandprops": end_o3_ratesandprops
}
```
But if these are expensive to construct and will add non-trivial import time this probably isn't the right way to go.

wbenoit26: Good stuff, I was hoping there was a better solution than this function. I'll try out both methods and see if there's a difference.

projects/sandbox/pyproject.toml
```diff
@@ -24,13 +24,13 @@ frame_type = "HOFT_C01"
 state_flag = "DCS-ANALYSIS_READY_C01:1"
 resnet = {layers = [3, 3, 3]}
 repository_directory = "${BASE_DIR}/model_repo/" 
-force_generation = false
```
alecgunny, line 27: I think we want this as `false` in general, but I understand why you would want it to be `true` while you were testing this. If the tests take a while to run, we can ignore fixing this and I can fix it in one of my PRs today, or if this would be a quick turnaround we can go ahead and change it here. Up to you.

alecgunny (approved):
Other than the one comment on the `pyproject.toml` this looks great, good work

projects/sandbox/pyproject.toml
wbenoit26: Currently on a flight and can't connect to LIGO, so let's fix it in one of your PRs

PR #185, opened 2022-11-01
use Segment and SegmentList features in generating glitches and background

deepchatterjeeligo:
No changes functionally. This patch makes use of intersection properties of Segments and its derivatives.


EthanMarx:
@deepchatterjeeligo Haven't forgotten about this. Ultimately I think these tools with eventually be migrated to [mldatafind](https://github.com/ML4GW/mldatafind) in some sort of vetoe manipulation package. Once we figure out how that is going to look, we can migrate this over.

PR #186, opened 2022-11-01
Fix glitch gen bug

### LINKED ISSUE #184: Pyomicron glitch generation condor issue
Utilizing `Pyomicron` to generate condor dags of omicron jobs  `sometimes` causes certain jobs to fail, and other times runs end to end with no problems.

EthanMarx:
Update `pyomicron` and `gwpy` in `glitch_generation` project to resolve condor dags not completing.
closes #184


alecgunny (changes requested):
Looks like there's a couple quick changes/commented lines in the `pyproject.toml` that just need to get swapped out then this should be good to go.

alecgunny (approved):

PR #188, opened 2022-11-02
Splitting up timeslide strain and timeslide output directories

alecgunny:
In order to easily reuse the same test data between runs, this splits out the timeslide strain data directory from where the directory where the outputs are saved. It also checks to see if a shift already has all of a particular segment's data generated, and will skip it if so. This makes it easy both to run the full pipeline via `pinto` without having to comment out or sit through this step, as well as to add shifts to an existing test set without having to regenerate everything.


alecgunny:
Also fixed issues with timestamp placement in vizapp event inspector

EthanMarx (approved):

Issue #152, opened 2022-09-04
Timeslide memory issue


EthanMarx:
Ran into the memory issue outlined [here](https://github.com/ML4GW/BBHNet/blob/main/projects/sandbox/timeslide_injections/timeslide_injections.py#L378L382) on a segment of length ~ 30,000 seconds. One solution off the top of my head would be to add a `max_segement_length` argument that ensures we won't run out of memory, and if we have segments larger than this, split them up.

EthanMarx:
One drawback of this method is these longer segments become discontiguous

alecgunny:
Closed by #189

PR #189, opened 2022-11-03
Timeslide memory issue

alecgunny:
Sampling waveforms one-at-a-time asynchronously rather than all at once to reduce memory overhead. Modularizing some of the components of the `timeslide_injections` project as well to simplify and clarify things. Still needs much more modularizing, but this could be the start of a much more general validation service.

Also updating the `ml4gw` reference to include corrections to individual IFO SNR calculation so that the SNRs reported in the vizapp are correct.


alecgunny:
Note that this overhaul changes where the `timeslide_injections` command points to, so you'll need to `pinto build -f` before running if you already have an existing environment

wbenoit26 (approved):

wbenoit26:
Closes #152

PR #190, opened 2022-11-04
Add Vetoe widget

EthanMarx:
Adds vetoe widget, and ability to adjust scatter plot attribute (e.g. detection statistic vs chirp mass) of injections in distribution plot. 

TODO:

- [x] Apply and save all combinations of vetoes upfront since calculating these every switch takes some time


EthanMarx:
Some enhancements that should probably be opened up as issues:

1. The Vetoe parsing and file discovery should probably be a component of mldatafind
2. Gating should be implemented at the strain level during timeslide generation

libs/analysis/bbhnet/analysis/distributions/cluster.py
```diff
@@ -99,3 +99,11 @@ def nb(self, threshold: Union[float, np.ndarray]):
         else:
             nb = np.sum(self.events >= threshold)
         return np.array(nb)
+
+    def copy(self):
```
alecgunny, line 103: Is there a reason to add this method here and to the `Foreground` object rather than using Python's built-in `copy.deepcopy` function?

EthanMarx: Nope. Wasn't familiar with it. Good to know, will adjust to the built in version.

projects/sandbox/vizapp/vizapp/plots/background.py
```diff
@@ -83,9 +87,9 @@ def configure_plots(self, height: int, width: int):
         )
         self.distribution_plot.add_layout(axis, "right")
 
-        r = self.distribution_plot.circle(
-            "detection_statistic",
-            "snr",
+        self.injection_renderer = self.distribution_plot.circle(
```
alecgunny, line 90: Small thing, but is there a reason this is saved as an attribute? Just curious

projects/sandbox/vizapp/vizapp/vetoes.py
```diff
@@ -0,0 +1,74 @@
...
+            [center - window - taper, center + window + taper]
+            for center, window, taper in zip(centers, windows, tapers)
+        ]
+    )
+
+    return vetoes
+
+
+@dataclass
+class VetoeParser:
```
alecgunny, line 32: Pedantic thing for here and elsewhere, but shouldn't the singular just be "veto"?

alecgunny (commented):
Only major sticking point for me is the `copy` method, everything else is mostly curiosity/aesthetic

projects/sandbox/vizapp/vizapp/plots/background.py
EthanMarx: Think this is just my bad habit of saving everything as attributes

projects/sandbox/vizapp/vizapp/app.py
```diff
@@ -54,13 +73,60 @@ def configure_widgets(self):
...
+                self.vetoed_distributions[vetoe_label] = {}
+                self.vetoed_foregrounds[vetoe_label] = {}
+                # calculate this vetoe combo for each norm and store
+                for norm, result in self.distributions.items():
+
+                    background = result.background.copy()
+                    for category in combo:
+
+                        vetoes = self.vetoe_parser.get_vetoes(category)
+                        background.apply_vetoes(**vetoes)
```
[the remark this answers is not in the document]
EthanMarx: It's not crazy expensive, but it definitely makes the app lag.

EthanMarx:
Was able to get the app running with the requested changes. Should be good to merge assuming tests pass. I assume you were you able to get it up and running @alecgunny

projects/sandbox/vizapp/vizapp/distributions.py
```diff
@@ -77,8 +77,8 @@ def get_foreground(
 
         with h5py.File(shift / "injection" / "params.h5") as f:
             event_times = f["geocent_time"][:]
-            h1_snrs = f["H1_snr"][:]
-            l1_snrs = f["L1_snr"][:]
+            h1_snrs = np.sqrt(f["H1_snr"][:])
```
alecgunny, line 80: Really quick: given that we're now using ML4GW with @deepchatterjeeligo 's fixes in it, this square rooting should actually be incorrect

alecgunny:
Building and running now, will confirm things are working once it's ready

projects/sandbox/vizapp/vizapp/distributions.py
EthanMarx: Done

alecgunny (approved):

PR #202, opened 2022-11-06
add power law distribution

deepchatterjeeligo:
Add amplitude distribution based on a volume distributed sources. This follows a power law with index -4. E.g.
image

### DIFF
--- a/libs/data/bbhnet/data/distributions.py
+++ b/libs/data/bbhnet/data/distributions.py
@@ -71,3 +71,40 @@ class LogNormal:
         if self.low is not None:
             x = torch.clip(x, self.low)
         return x
+
+
+class PowerLaw:
+    """
+    Sample from a power law distribution,
+    .. math::
+        p(x) \approx x^{-\alpha}.
+
+    Index alpha must be greater than 1.
+    This could be used, for example, as a universal distribution of
+    signal-to-noise ratios (SNRs) from uniformly volume distributed
+    sources
+    .. math::
+
+       p(\rho) = 3*\rho_0^3 / \rho^4
+
+    where :math:`\rho_0` is a representative minimum SNR
+    considered for detection. See, for example,
+    `Schutz (2011) <https://arxiv.org/abs/1102.5421>`_.
+    """
+
+    def __init__(
+        self, x_min: float, x_max: float = float("inf"), alpha: float = 2
+    ) -> None:
+        self.x_min = x_min
+        self.x_max = x_max
+        self.alpha = alpha
+
+        self.normalization = x_min ** (-self.alpha + 1)
+        self.normalization -= x_max ** (-self.alpha + 1)
+
+    def __call__(self, N: int) -> torch.Tensor:
+        u = torch.rand(N)
+        u *= self.normalization
+        samples = self.x_min ** (-self.alpha + 1) - u
+        samples = torch.pow(samples, -1.0 / (self.alpha - 1))
+        return samples
--- a/libs/data/tests/test_distributions.py
+++ b/libs/data/tests/test_distributions.py
@@ -1,5 +1,9 @@
 from math import pi
 
+import numpy as np
+import pytest
+from scipy import optimize
+
 from bbhnet.data import distributions
 
 # TODO: for all tests, how to validate that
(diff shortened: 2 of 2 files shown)


libs/data/bbhnet/data/distributions.py
```diff
@@ -71,3 +71,26 @@ def __call__(self, N: int) -> torch.Tensor:
         if self.low is not None:
             x = torch.clip(x, self.low)
         return x
+
+
+class VolumeDistributedSNR:
```
alecgunny, line 76: For the sake of consistency, it might make sense to have the name of this class describe the distribution itself, e.g. `PowerLaw` or something. That SNR distributions tend to follow it is a useful application, but not necessarily a defining attribute of the distribution.

libs/data/bbhnet/data/distributions.py
```diff
@@ -71,3 +71,26 @@ def __call__(self, N: int) -> torch.Tensor:
...
+    .. math::
+
+       p(\rho) = 3*\rho_0^3 / \rho^4
+
+    where :math:`\rho_0` is a representative minimum SNR
+    considered for detection. See, for example,
+    `Schutz (2011) <https://arxiv.org/abs/1102.5421>`_.
+    """
+
+    def __init__(self, rho_0: float = 4) -> None:
```
alecgunny, line 90: It might be worth doing a couple things here, curious to get your thoughts @EthanMarx , especially in light of your proposal to generalize this submodule to `ml4gw`:
1. Use a more descriptive kwarg for the minimum value of the distribution, e.g. `min_val` or `low`
2. Don't give this a default value: it's a sufficiently defining property of the distribution that we should make users be explicit about the value they want.
3. Add the power as an argument (I'm fine having this default to 4), and just raise to the `-1 / (pow - 1)` at call time. This allows this both to be more general, as well as allows us to pass a torch scalar to this value that we can update over time, in case users want to manipulate their distribution through the course of training (e.g. start this off at 4, then have it slowly increase over time to focus on lower SNR events)

alecgunny (commented):
Couple points for discussion in the changes, but overall looks good

libs/data/bbhnet/data/distributions.py
EthanMarx: Agree on all these points. Power laws with varying indices will certainly have other use cases.

deepchatterjeeligo: easy generalization. will do.

libs/data/tests/test_distributions.py
```diff
@@ -57,3 +61,19 @@ def test_log_normal():
...
+    """Test VolumeDistributedSNR against expected ideal distribution"""
+    ref_snr = 8
+    sampler = distributions.VolumeDistributedSNR(ref_snr)
+    samples = sampler(10000).numpy()
+    # check 1/rho^4 behavior
+    counts, ebins = np.histogram(samples, bins=100)
+    bins = ebins[1:] + ebins[:-1]
+    bins *= 0.5
+
+    foo = lambda x, a, b: a * x ** (-b)  # noqa E731
```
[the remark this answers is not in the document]
deepchatterjeeligo: I find this neater. Shall we consider this an exception, since its a test script anyway.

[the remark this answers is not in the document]
alecgunny: Given that it's just a test script and that it's obvious that this function won't raise an error, I would say keeping it is fine for now. But in general in the future I think the PEP [programming guidelines](https://peps.python.org/pep-0008/#programming-recommendations) make a good case why this syntax is worth avoiding. Namely

> The (`def`) form means that the name of the resulting function object is specifically (‘foo’) instead of the generic ‘<lambda>’. This is more useful for tracebacks and string representations in general. The use of the assignment statement eliminates the sole benefit a lambda expression can offer over an explicit def statement (i.e. that it can be embedded inside a larger expression)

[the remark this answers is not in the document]
deepchatterjeeligo: ok.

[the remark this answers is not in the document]
alecgunny: Huh ok I did not realize that `black` didn't allow this one-line `def` syntax. Kind of weird frankly but good to know, still probably more clear but I get the appeal of the lambda formulation now

libs/data/bbhnet/data/distributions.py
alecgunny: Last comment: looks great and I like the addition of a max value, but would it be sensible to have this default to `float("inf")`. I have to imagine that will cover >99% of use cases

libs/data/tests/test_distributions.py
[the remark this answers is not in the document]
deepchatterjeeligo: nevermind. I used the PEP syntax.

libs/data/bbhnet/data/distributions.py
deepchatterjeeligo: OK

deepchatterjeeligo: Actually, another case where I think this can be used is the mass distribution for BBHs. And there, there is a cut off around 100 solar masses. So thinking about that, do we still want a default? I just kept the index default at 2, since its the integer after 1, no good reason apart from that.

deepchatterjeeligo: I mean default of infinity as the upper limit

alecgunny: I'm still inclined to say set the default upper limit at infinity for a couple reasons:
1. My impression is that most use cases won't have a sensible upper limit to set, and are happy to rely on the statistics to functionally disallow really high values. This won't always be the case, but that's why they're defaults and not hard-coded values. But maybe I'm thinking too generally and in GW there will more often than not be a sensible max to set.
2. If (1) is true even 51% of the time, I'm inclined to set the default just because of how verbose `float("inf")` is and that it's not necessarily a super well-known bit of Python syntax (I would bet quite a few people will unnecessarily `import numpy` to use `np.inf`).

deepchatterjeeligo: ok.

alecgunny (approved):
Looks great!

PR #206, opened 2022-11-08
Larger batch inference

### LINKED ISSUE #200: Larger batch inference
As ML4GW/hermes#35 outlines, we can, with very little effort, be doing inference with much larger batch sizes, which should drastically increase our throughput.

### LINKED ISSUE #204: Use latest `hermes.quiver` and `clean = true` in `export` project
Now that ML4GW/hermes#34 has been merged, passing `clean=True` to a new model repo shouldn't raise an error. There seems to be an issue when re-running export on an already initialized repository when this isn't set. This should be fixed separately (e.g. if you actually want to export a newer version), but in the meantime to get things working we should use the latest version of `hermes` and set `clean=True` by default to get repeated runs of the same pipeline functional while we're developing.

alecgunny:
Using latest version of `hermes` to support large batch inference, and making some adjustments to inference and export projects to optimize performance on a per-gpu basis. Closes #200 , closes #204

### DIFF
--- a/projects/sandbox/infer/infer/main.py
+++ b/projects/sandbox/infer/infer/main.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import time
 from pathlib import Path
 from typing import Iterable, Optional
@@ -25,9 +26,9 @@ def main(
     fields: Iterable[str],
     sample_rate: float,
     inference_sampling_rate: float,
-    inference_rate: float,
     batch_size: int,
     num_workers: int,
+    throughput_per_gpu: float,
     streams_per_gpu: int,
     model_version: int = -1,
     max_seconds: float = 131072,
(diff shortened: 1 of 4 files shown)


projects/sandbox/infer/infer/main.py
```diff
@@ -52,12 +54,13 @@ def main(
             write_dir,
             fields,
             client,
-            stride_size=stride_size,
+            inference_sampling_rate=inference_sampling_rate,
+            sample_rate=sample_rate,
             batch_size=batch_size,
             fduration=fduration,
-            inference_rate=inference_rate,
+            throughput=throughput_per_gpu * num_gpus,
```
EthanMarx, line 61: What exactly is setting the throughput doing?

EthanMarx (approved):
Looks good. Do you have a good idea for the optimal `throughput` value, or does that still have to be experimented with?

PR #221, opened 2022-11-17
Validation improvements

### LINKED ISSUE #41: Which validation metrics to use
One of the bigger issues we face is the impact of low-probability, high-SNR glitches increasing our false alarm rate. While we want to be monitoring our performance on the "true" data distribution during training, it also might be helpful to evaluate our efficacy on these events which don't occur frequently enough to allow a meaningful sample size from the lengths of background data we can reasonably do validation on.

One solution might be to create a dataset of _only_ high (for some definition of "high") SNR glitches that gets iterated through sequentially during a separate validation step. For the data-loading portion, it might be as simple as rejiggering the `GlitchSampler` to be able to sample deterministically and sequentially.
- What does "sequentially" mean here? Do we sample each kernel (or sample with some stride) that contains the glitch "trigger"? Or just produce one kernel per glitch?
- How does the answer to that question impact how we evaluate our performance? Kernels from the same glitch aren't independent, do we instead do some sort of accumulation across these kernels to mimic the matched filtering inference approach? And compare against what, a similar analysis of the validation injections?

### LINKED ISSUE #170: Improve validation during training

alecgunny:
Implement measurements of recall (fraction of signals recovered) vs. background and glitch datasets separately. Closes #41 , closes #170 .

For background, outputs are pooled across adjustable windows at an adjustable stride to simulate test-time clustering, and recall is measured using thresholds from the top `k` background "events".

For glitches, recall is measured using thresholds from adjustable specificity levels (fraction of glitches correctly classified at a given threshold).

Monitoring (for early stopping and model selection) can be performed against any threshold for either metric. Checkpointing at a specified cadence has been added as well.

- [ ] Add `Recorder` and `Validator` parameters to `train:main` function definition
- [ ] Add tests
- [ ] Fix `sample_rate` used for background pooling to be `1 / stride`


alecgunny:
Ok I have screwed something up royally with the commit ordering here, I may need to rework this

PR #223, opened 2022-11-17
Validation fixes

### LINKED ISSUE #41: Which validation metrics to use

### LINKED ISSUE #170: Improve validation during training

alecgunny:
Implement measurements of recall (fraction of signals recovered) vs. background and glitch datasets separately. Closes #41 , closes #170 . Re-creation of #221 but with the git history better aligned.

For background, outputs are pooled across adjustable windows at an adjustable stride to simulate test-time clustering, and recall is measured using thresholds from the top `k` background "events".

For glitches, recall is measured using thresholds from adjustable specificity levels (fraction of glitches correctly classified at a given threshold).

Monitoring (for early stopping and model selection) can be performed against any threshold for either metric. Checkpointing at a specified cadence has been added as well.

- [x] Add `Recorder` and `Validator` parameters to `train:main` function definition
- [x] Add tests
- [x] Fix `sample_rate` used for background pooling to be `1 / stride`

### DIFF
--- a/projects/sandbox/train/tests/test_validation.py
+++ b/projects/sandbox/train/tests/test_validation.py
@@ -1,0 +1,80 @@
+import pytest
+import torch
+from train import validation
+
+
+def test_background_recall():
+    metric = validation.BackgroundRecall(10, 10, 3)
+    background = torch.zeros((31,))
+
+    # after pooling, the top 3 values should come
+    # out to 2, 1.5, and 0.8
+    background[3] = 1
+    background[4] = 1.5
+    background[10] = 2
+    background[14] = 0.5
+    background[18] = 0.9
+    background[22] = 0.8
+    background[30] = 3  # this will get trimmed
+
+    signal = torch.arange(0.5, 2.5, 0.1)
+    scores = metric(background, None, signal).cpu().numpy()
+    assert len(scores) == 3
+    assert all([i == j for i, j in zip(scores, metric.values)])
+
+    assert pytest.approx(scores[0], 0.25)
+    assert pytest.approx(scores[1], 0.5)
+    assert pytest.approx(scores[2], 0.85)
+
+
+def test_glitch_recall():
+    metric = validation.GlitchRecall([0.5, 0.75, 1])
+    glitches = torch.arange(11).type(torch.float32)
+    signal = torch.arange(4, 12).type(torch.float32)
+
+    scores = metric(None, glitches, signal).cpu().numpy()
+    assert len(scores) == 3
+    assert all([i == j for i, j in zip(scores, metric.values)])
+
+    assert pytest.approx(scores[0], 7 / 8)
+    assert pytest.approx(scores[1], 4 / 8)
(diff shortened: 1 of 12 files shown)


alecgunny:
Tests aren't exhaustive, but cover the most relevant functionality.

projects/sandbox/train/tests/test_validation.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # after pooling, the top 3 values should come
+    # out to 2, 1.5, and 0.8
+    background[3] = 1
+    background[4] = 1.5
+    background[10] = 2
+    background[14] = 0.5
+    background[18] = 0.9
+    background[22] = 0.8
+    background[30] = 3  # this will get trimmed
```
EthanMarx, line 18: is this a `torch.nn.functional.maxpool1d` artifact?

EthanMarx (approved):

projects/sandbox/train/train/train.py
```diff
@@ -204,9 +234,56 @@ def main(
     # into one file for simplicity
     background = load_background(hanford_background, livingston_background)
     if valid_frac is not None:
-        background, valid_background = split(background, 1 - valid_frac, 1)
-        valid_injector.fit(H1=background[0], L1=background[1])
-        valid_waveforms, _ = valid_injector.sample(-1)
+        # split up our background data into train and validation splits
+        background, valid_background = split(background, 1 - valid_frac, -1)
```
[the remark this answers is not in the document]
alecgunny: Yeah in this case the background array is 2D so those should be equivalent. I don't know exactly why I opted for this change, but now that I'm thinking about it I do like the idea of `-1` as a handy shorthand in our context for "do this operation along the time dimension," since that's where it will be regardless of the shape of the array

projects/sandbox/train/tests/test_validation.py
alecgunny: Yeah exactly, actually didn't know this until I built this but I guess it adopts the convention of sloughing data off the end that can't be fit into a full kernel rather than doing the op on a shorter kernel. Good to keep in mind for the future I guess

PR #226, opened 2022-11-21
Splitting segments

EthanMarx:
Adds optional `chunk_length` parameter to `timeslide_injections` project. If passed, will chunk segments longer than `chunk_length` to avoid loading too much data into memory.


alecgunny:
It looks like you have a lot of extra commits in here because your local `main`'s history has extra commits that the upstream `main` doesn't have (likely because of extra merge commits when you've done `git pull upstream main`). You might want to consider doing something like (assuming you're currently in this branch):
```
git branch -d main  # delete local main
git fetch upstream
git checkout -b main --track upstream/main  # reset local main to have same history as remote
git checkout -b splitting-segments-rebased
git cherry-pick 330f04b 8318eef 9e76463  # apply the three commits from this PR on top of main
```

EthanMarx:
> 



> It looks like you have a lot of extra commits in here because your local `main`'s history has extra commits that the upstream `main` doesn't have (likely because of extra merge commits when you've done `git pull upstream main`). You might want to consider doing something like (assuming you're currently in this branch):
> 
> ```
> git branch -d main  # delete local main
> git fetch upstream
> git checkout -b main --track upstream/main  # reset local main to have same history as remote
> git checkout -b splitting-segments-rebased
> git cherry-pick 330f04b 8318eef 9e76463  # apply the three commits from this PR on top of main
> ```

Ah weird. Was there anything I did wrong? Haven't run into this before.

alecgunny:
I think we all have and just maybe haven't noticed it (I just had this on a PR last week too), but I think at some point our local `main` branches branched slightly from the upstream `main`. Whenever this happened, the next time we tried to `git pull upstream main`  to pull in the changes from e.g. the latest PR, Git created a local merge commit that joined our local version of `main` with the one we just pulled down to signify the rejoining of these two branches.

The problem is from then on out, now you've added _another_ commit to your local `main` that is divergent from the upstream (which doesn't have that merge commit), and the problem just propagates. So every time you pull the remote `main` down, it's reconciling your history with the upstream one (which is missing commits in its view), and creating a merge commit.

The way to get around this is to start doing `git pull --rebase upstream main`. This way if for some reason your histories are different (and ideally they shouldn't be), git will try to replay your local changes on top of `main`. This will still leave extra commits in the history, but if there's anything problematic this will just cause an error and you can fix your history or notice that you've done something wrong and just do a fresh fetch.

You can also set this to be the default behavior for just this branch by setting
```
[branch "main"]
  rebase = true
```
in  your local `.git/config`

PR #229, opened 2022-11-23
Train project tests

wbenoit26:


projects/sandbox/train/pyproject.toml
```diff
@@ -9,8 +9,9 @@ license = "MIT"
 [tool.poetry.dependencies]
 python = ">=3.8,<3.11"
 "bbhnet.data" = {path = "../../../libs/data", develop  = true}
-"bbhnet.trainer" = {path = "../../../libs/trainer", develop = true}
+"bbhnet.io" = {path = "../../../libs/io", develop = true}
```
alecgunny, line 12: Can we remove the `io` lib dependency now that the tests are updated?

alecgunny (changes requested):
I think just removing that dep and re-locking should do it

projects/sandbox/train/pyproject.toml
wbenoit26: Oh, sorry, I meant to move this back to a draft PR. I'm still figuring out how to check on the content of `validator`.

wbenoit26: I've got this updated in my local version.

wbenoit26:
@alecgunny I tried to make that signal comparison test we discussed earlier. For the life of me, I can't see why it's not working. I figured I'd put it here to get a second set of eyes on it, and to see if you had any comments on the other tests that I added.

Oh, and it looks like making and un-making changes in `validation.py` created another large diff - not sure why, but I did check, and nothing is actually different.

alecgunny (approved):

PR #231, opened 2022-11-27
Incorporate mldatafind

EthanMarx:
Incorporates [mldatafind](https://github.com/ML4GW/mldatafind) library into `timeslide`, `train` and `background` projects. 

Some comments:

- Omicron runs over training and testing datasets are now split into separate directories to reduce any confusion

- mldatafind [`authenticate`](https://github.com/ML4GW/mldatafind/blob/main/mldatafind/authenticate.py#L8) function is now called at the beginning of the `background` project

- `channels` and `state_flags` are now passed as a list, one for each `ifo`. This will allow us to include Virgo and other detectors in the future as channel names ( and `state_flags`) differ from the LIGO detectors. 

- We don't fully take advantage of the `find_data` generator in the `timeslide` project. This would require a bit of refactoring that I didn't want to attack in this PR


EthanMarx:
Bit off more than I can chew here so refactoring into smaller PR's

PR #232, opened 2022-12-02
Prior and datagen documentation

wbenoit26:
Creating a single PR for the `datagen` project and removing the other one that I made


wbenoit26:
I find this commit history suspect - I'm going to redo my changes with a clean branch.

PR #236, opened 2022-12-05
Adding normalizer check and accelerating distribution clustering

### LINKED ISSUE #235: Check for `nan` during normalization
If a model gets trained poorly and outputs constant or nearly constant predictions, normalization during analysis will lead to division by zero and therefore `nan` values. This becomes problematic because comparisons on `nan`s will always return `False`, so [this line](https://github.com/ML4GW/BBHNet/blob/c64c1183971110769ca3e06a167958b1f5a29eae/libs/analysis/bbhnet/analysis/distributions/cluster.py#L79) in `bbhnet.analysis.distributions.ClusterDistribution` will always return `False` and we end up with an order of magnitude more events than expected.

Instead, we should just raise an error in `bbhnet.analysis.normalizers.GaussianNormalizer` if `(self.scales == 0).any()`

alecgunny:
Closes #235 and speeds up clustering by a factor of ~20 by jumping peak values rather than doing comparisons for every value in the array

### DIFF
--- a/libs/analysis/bbhnet/analysis/distributions/cluster.py
+++ b/libs/analysis/bbhnet/analysis/distributions/cluster.py
@@ -61,26 +61,23 @@ class ClusterDistribution(Distribution):
         # update background time
         self.Tb += t[-1] - t[0] + t[1] - t[0]
 
-        remove_indices = []
-
         # infer number of samples in half window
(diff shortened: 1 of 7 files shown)


alecgunny:
Agreed, just pushed different wording. Let me know how that looks

wbenoit26 (approved):

PR #241, opened 2022-12-09
add glitch start to config

EthanMarx:
Adds `glitch_start` to `pyproject.toml` base, and uses it for start of glitch generation. Renames `start` and `stop` parameters to `train_start` and `train_stop` for clarity.


alecgunny (approved):
Will this cause any issues when we try to fetch the actual glitch data if it's really big?

EthanMarx:
You mean when loading in the triggers?  I don't think a days worth will cause issues, especially with the clustering we do, but good to keep in mind.

PR #243, opened 2022-12-15
Sample priors from file

### LINKED ISSUE #217: Enable parameter sampling from data files
For even more realistic testing data, we may want to sample parameters from distributions inferred from gravitational wave detections, which typically aren't well-described by analytical functions (see [here](https://dcc.ligo.org/DocDB/0180/T2100512/009/README.html)).

Adding the functionality to sample from numerical distributions, or a mix of a bilby prior dictionary and a numerical distribution, would let us use this more realistic data.

### LINKED ISSUE #239: Can't pickle end_o3_ratesandpops prior
This is more an FYI than something we need to fix, but using the `end_o3_ratesandpops` prior in `generate-timeslides` raises an error because it can't pickle the conditional function inside the prior function.

As noted in #217, it would be better to move to a numerical prior anyway, which would remove this issue. For the time being, we can just not use this prior.

wbenoit26:
Resolves #217, resolves #239


alecgunny (commented):
See note about implementation feedback. Where did we land on the independent sampling functionality?

wbenoit26:
There are two versions of the model, one in which the masses are independent of each other, and one in which the joint mass pdf is multiplied by a power law on the mass ratio. All other parameters are independent of each other.

projects/sandbox/datagen/datagen/utils/injection.py
```diff
@@ -1,11 +1,52 @@
...
+from typing import Callable, Dict, List, Optional, Tuple
 
+import h5py
 import numpy as np
 from bilby.gw.conversion import convert_to_lal_binary_black_hole_parameters
 from bilby.gw.source import lal_binary_black_hole
 from bilby.gw.waveform_generator import WaveformGenerator
 
 
+def sample_params(
```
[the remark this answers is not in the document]
wbenoit26: I think bilby's [interpolated prior](https://git.ligo.org/lscsoft/bilby/-/blob/master/bilby/core/prior/interpolated.py#L8) more or less does what we want here, though with a different implementation. We'll still need to build something to deal with that particular parameter file, though.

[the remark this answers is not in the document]
alecgunny: Oh woah ok nice that's super cool, does the parameter file contain any information about `p(x)` to feed to the `yy` argument though?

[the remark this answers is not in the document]
wbenoit26: It looks like there's actually an associated file of the pdf for each parameter, though it wouldn't be hard to make a function to compute that ourselves if we wanted something more general.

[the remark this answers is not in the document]
alecgunny: Hmm that's interesting, how is the pdf specified? Is it the `p(x)` for each value in the associated file, or over some lin/logspaced range of input values? If it's the latter, we could just use this to pass to the interpolated prior and drop the other file completely (though maybe this is getting away from what the point of the parameter file was to begin with)

EthanMarx:
Been thinking about how this can nicely fit in with #251. From an importance sampling perspective, we'll need to be able to calculate the probability of a given parameter sample. All of the `bilby.prior` objects (including the interpolated) have a `.prob` method that does this. With this in mind, I think we should use the `bilby.core.prior.Interped` in favor of passing arrays of parameters. In principle, these parameters should have been drawn from something like a `bilby.core.prior.Interped` prior anyways.

wbenoit26:
I fully agree. I'll try to get this PR ready to go in the next few days.

EthanMarx:
Awesome, I've fleshed out what the importance sampling might look like in #251. If we decide to create a wrapper or subclass around `bilby.Prior`, as long as it implements a `.prob `method, it should be compatible

projects/sandbox/datagen/datagen/utils/priors.py
```diff
@@ -1,109 +1,131 @@
...
+            grid, pdf = pdf_from_events(events[name])
+            prior[name] = Interped(grid, pdf, np.min(grid), np.max(grid))
+
+    return prior
+
+
+def assign_prior_names(prior: BBHPriorDict) -> BBHPriorDict:
+    for key in prior.keys():
+        if hasattr(prior[key], "name"):
+            print("check")
```
alecgunny: Assume this is a leftover debug statement

projects/sandbox/datagen/datagen/utils/priors.py
```diff
@@ -1,109 +1,131 @@
...
+    return prior
+
+
+def uniform_extrinsic():
+    prior = BBHPriorDict()
+    prior["dec"] = Cosine()
+    prior["ra"] = Uniform(0, 2 * np.pi)
+    prior["theta_jn"] = 0
+    prior["phase"] = 0
+    prior = assign_prior_names(prior)
```
alecgunny: I really like this functionality, I've always thought the naming and key assignment was redundant. Out of curiosity, what happens if the parameters don't have names?

projects/sandbox/datagen/datagen/utils/priors.py
```diff
@@ -1,109 +1,131 @@
...
+    prior["phi_12"] = Uniform(0, 2 * np.pi)
+    prior["phi_jl"] = 0
+    prior = assign_prior_names(prior)
+
+    return prior
+
+
+def power_law_dip_break():
+    prior = uniform_extrinsic()
+    event_file = "/home/william.benoit/\
```
alecgunny, line 143: Is there a way to do this without hard-coding? How would this look with this as an argument?

projects/sandbox/datagen/datagen/utils/timeslides.py
```diff
@@ -24,10 +24,13 @@ def __init__(
         waveform_duration: float,
         max_shift: float,
         jitter: float,
-        parameter_file: Optional[Path] = None,
+        buffer: float = 0,
+        spacing: float = 0,
     ) -> None:
         self.prior = prior
```
alecgunny, line 30: Would it make sense to call `prior()` here rather than constructing the dict each time?

projects/sandbox/datagen/datagen/utils/priors.py
```diff
@@ -1,109 +1,131 @@
...
+        O1O2O3all_mass_h_iid_mag_iid_tilt_powerlaw_redshift_maxP_events_all.h5"
+    prior |= read_priors_from_file(event_file)
```
alecgunny, line 145: I didn't know this fun little piece of syntax, rad!

alecgunny (changes requested):
This looks really good, only major things are the print statement and the calling of `prior()` during `Sampler.__init__`, though curious to get your and @EthanMarx 's thoughts on how to handle the filename. Even if it makes things too complicated to use an arg, maybe an environment variable that defaults to that value?

projects/sandbox/datagen/datagen/utils/timeslides.py
alecgunny: (As in constructing it by calling `prior()` in each `__call__`)

projects/sandbox/datagen/datagen/utils/priors.py
wbenoit26: Yeah, I think you're looking at an older commit. Should be gone in the most recent one.

projects/sandbox/datagen/datagen/utils/priors.py
wbenoit26: Agreed. I actually moved away from using this function in the latest commit because most of the priors don't need names. For the two that do, I've hardcoded the names in.

projects/sandbox/datagen/datagen/utils/timeslides.py
wbenoit26: Yes, agreed.

projects/sandbox/datagen/datagen/utils/priors.py
wbenoit26: I'm open to ideas, but I do think having the file name as an argument would be overkill. I would rather each prior function correspond to a specific set of priors than have a generic `read_from_file_prior()`, just for the sake of clarity. Like, this prior corresponds to this file, and that's all it does every time. This way, anyone who wants to use a particular prior doesn't have to wonder which file they need.

projects/sandbox/datagen/datagen/utils/priors.py
EthanMarx: I *think* the names are used for bilbys plotting functionality (i.e. they are optional). Will, which priors need names, and do you know why?

wbenoit26: Generally, anything that inherits from bilby's `Cosmological` class requires the name to be specified, and to be in `["redshift", "luminosity_distance", "comoving_distance"]`. For us, that means we need to specify names at initialization for `UniformComovingVolume` and `UniformSourceFrame`.

EthanMarx:
With regards to the drawing parameters from a file, I am pretty certain those samples were drawn using [`gwpopulation`](https://colmtalbot.github.io/gwpopulation/)  (specifically, I think [this](https://colmtalbot.github.io/gwpopulation/_autosummary/gwpopulation.models.mass.double_power_law_peak_primary_mass.html#gwpopulation.models.mass.double_power_law_peak_primary_mass) is the model you are referring to). If we think performing injections with these more complex models is useful, we should probably use those priors directly to get a denser sampling of the space.  

Was implementing this prior motivated by doing the VT calc? Is this different than the O3 rates and pop prior?

wbenoit26:
They were, from [this](https://github.com/afarah18/gwpopulation/blob/acdca206d7d786de596307d0d98867e290d63249/gwpopulation/models/mass.py#L211) model. I agree that using those priors directly would be ideal, but I'm worried about how much complexity that might introduce for something that we don't need to be exact on. Though, I saw that you were thinking of refactoring the prior code into its own library again, so maybe that could play into it.

No, I was working on this before the VT calculation came up, and it is different from the O3 rates and pops prior. This implements a more complex prior on the masses.

EthanMarx:
> They were, from [this](https://github.com/afarah18/gwpopulation/blob/acdca206d7d786de596307d0d98867e290d63249/gwpopulation/models/mass.py#L211) model. I agree that using those priors directly would be ideal, but I'm worried about how much complexity that might introduce for something that we don't need to be exact on. Though, I saw that you were thinking of refactoring the prior code into its own library again, so maybe that could play into it.
> 
> No, I was working on this before the VT calculation came up, and it is different from the O3 rates and pops prior. This implements a more complex prior on the masses.



> They were, from [this](https://github.com/afarah18/gwpopulation/blob/acdca206d7d786de596307d0d98867e290d63249/gwpopulation/models/mass.py#L211) model. I agree that using those priors directly would be ideal, but I'm worried about how much complexity that might introduce for something that we don't need to be exact on. Though, I saw that you were thinking of refactoring the prior code into its own library again, so maybe that could play into it.
> 
> No, I was working on this before the VT calculation came up, and it is different from the O3 rates and pops prior. This implements a more complex prior on the masses.

Got it. I guess what I'm confused about is why we need this prior (which from my understanding is mostly used for performing population analyses) if we don't need something that is that 'exact'.

wbenoit26:
Well, I suppose it depends on how we define "exact". We want to be testing on injections that are drawn from the true population distribution, but it probably doesn't matter if we're just a little bit off (and estimating the distributions this way does get us pretty close). None of the previous sets of priors were at all close to this model, and I think someone could reasonably ask about them if they were concerned about the validity of our results.

EthanMarx:
I see. I have all the same concerns. This seems to boil down to what a good performance metric is. I think the most important criteria is that our metric can be compared with matched filter pipelines. I think a VT using our best guess of the true astrophysical population is a useful metric, and hopefully other pipelines have a number we can benchmark against.

What I gathered from our discussion with Reed was that we can inject a somewhat reasonable distribution (i.e. the end_o3_rates_and_pop), and through importance sampling, be able to calculate VT's for other arbitrary distributions (e.g, at specific points in the mass space). In other words, we don't need to go through the trouble of crafting the perfect distribution. 

If we think using population informed priors is generally useful, it might be worth investing time into figuring out how we can fit `gwpopulation` into our prior capabilities. We can open that up as an issue for later if its too much work for now.

Last thing: I think Alecs concern with the hardcoded file name is that then this prior is only available if a user runs the pipeline on CIT (or wherever you stored that file). If we don't want to go down the `gwpopulation` route, one solution would be to store that file in the repo itself.

wbenoit26:
Ah, that's true. Still, I think there's no reason not to start with our best guess of the population distribution. Or, at least we would want this prior so we could do the VT calculation for this distribution, right? 

Augmenting our current prior setup with `gwpopulation` seems like more work than we need to be doing right now - at this point, I think we just want something that works.

Yeah, I see the concern there. I would've added the file to the repo originally, but it's a little big, ~60MB. Though as I type this, I realize that I could make a separate file with just the BBH events, which would be ~10% of the size.

projects/sandbox/datagen/datagen/utils/priors.py
alecgunny: Disregard, just seeing the discussion about this below

wbenoit26:
@alecgunny Is this good to go?

alecgunny:
@wbenoit26 looks good to me sorry about that, didn't know if there was some resolution to the above conversation achieved during the sync last week. @EthanMarx I'm going to give this a thumbs up but feel free to chime in here or raise a new issue with any lingering feedback

alecgunny (approved):

EthanMarx:
Yeah should merge this. I'll open up some issues for the improvements

PR #244, opened 2022-12-15
Tensorrt update

alecgunny:
Updating TensorRT for functionality on Ampere GPUs. Won't be functional until a new container image for Triton 22.07-py3 is made available on LDG which has the appropriate TensorRT version
- [x] ML4GW/hermes#13

### DIFF
--- a/libs/architectures/poetry.lock
+++ b/libs/architectures/poetry.lock
@@ -220,9 +220,9 @@ igwn-auth-utils = ">=0.3.1"
 ligo-segments = ">=1.0.0"
 
 [package.extras]
-docs = ["sphinx", "sphinx-automodapi", "sphinx-immaterial"]
-lint = ["flake8 (>=3.7.0)", "flake8-bandit"]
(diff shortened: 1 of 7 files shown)


wbenoit26:
When you say "won't be functional", does that mean that nothing will change until that happens, or that things will break if this change is merged now?

alecgunny:
Things would break since the version of Triton that's sitting in a container on LDG would be running a version of TensorRT (8.2.3.1 I believe) that doesn't match this version and NVIDIA, in their infinite wisdom, insists that the version of TensorRT used to compile a model and the one used to a run a model need to match _exactly_, down to the most minor version number.

alecgunny:
Waiting on opensciencegrid/cvmfs-singularity-sync#404 to get merged

alecgunny:
Good to go once ML4GW/hermes#73 gets merged and this updates the hermes ref to that version

alecgunny:
Tested export and inference on both V100 and A30 GPUs

wbenoit26 (approved):

PR #245, opened 2022-12-15
Normalizing line endings

alecgunny:
Normalizing all line endings to be Unix style and adding .gitattributes config to enforce this going forward


alecgunny:
Going to request a similar pass on the failing test for now just because fixing it in a separate branch and then replaying this over that will be more complicated than it's worth.

For anyone who wants to rebase a PR on top of this once it's merged, just use `git rebase -i main`, which will tell `rebase` to ignore whitespace differences

EthanMarx (approved):

Issue #198, opened 2022-11-06
Larger glitch dataset/generation issues


alecgunny:
@EthanMarx what's the status of some of the issues we're seeing with omicron? If/when those get resolved, how large of glitch datasets will we be working with, and what are the easiest strategies for growing those?

EthanMarx:
The issue appears to be resolved by upgrading to the latest omicron version (3.2.0, that is now stable). I would like both you and Will to run with this new environment so we can cross check results. I'm seeing ~ 700 glitches for each IFO during the training period. 

A few ideas for generating larger glitch datasets:

1. We currently restrict all of our analysis to coincident times, and thus don't take advantage of times where only 1 IFO is online. It might be useful to think of a strategy to incorporate glitches from these times as well. 
2. Implement some bootstrapping method where we incorporate events our network performed poorly on back into the glitch dataset. 
3. Increase the length of our training period

These are the only ways I can think of to get more "natural" glitches. 

I think the major problem with our glitch dataset is a lack of high SNR glitches. One thing that would be super interesting would be to do some sort of SNR re-weighting with glitches. The problem here is we would need a method to extract the glitch from the background, otherwise we would be raising the noise floor as well. We could also think about sampling the glitches we currently have from some SNR distribution, so the network sees the louder glitches more often.

alecgunny:
Ran this morning with the newer version of omicron and am seeing 195 Hanford glitches and 751 Livingston glitches.

Would it make sense to pull from periods before the training period for glitches? Most of the reason we keep the training period relatively short is so that we can keep all the data in memory at once. We'll probably eventually relax this constraint, but until then we should be free to pull glitches from any time we want before the test period right?

EthanMarx:
This makes a lot of sense. In an online scenario we'd have access to those glitches anyways, so I don't see why not.

EthanMarx:
I think I've finally figured out the issue here, and for the moment it might be out of our control. 

To summarize the problem we were seeing:

1. Omicron jobs would randomly fail due to some error loading the frame cache file.
2. Which jobs failed was more or less random: 2 different workflow runs might have different jobs failing, or no jobs failing at all. 

In addition to the condor dag file, pyomicron creates a bash script that contains all the same jobs. When running this script, the jobs would consistently complete without failure. The major difference between running via condor and simply running the bash script is that the bash script jobs are run serially, while the condor jobs are launched in parallel. 

[...]

PR #248, opened 2022-12-28
parallelize glitch project

EthanMarx:
1. Separates training and testing set runs into separate directories for ease of mind
2. Utilizes mldatafind `find_data` when querying strain
3. Adds `analyze_testing_set` flag 
4. Parallelizes pyomicron condor dag generation jobs using a `ProcessPool`
5. Allows querying glitches from multiple segments (i.e. distinct trigger files) for larger datasets


alecgunny:
Just to make sure, this is still subject to the issues outlined in #198 ? (Not that that would hold this up, since this is valuable in its own right, just clarifying)

EthanMarx:
> Just to make sure, this is still subject to the issues outlined in #198 ? (Not that that would hold this up, since this is valuable in its own right, just clarifying)

Unfortunately yes. Will have to wait to hear back from omicron developers.

alecgunny:
Sounds good, this is great either way

alecgunny (approved):
The one thing I'll say is that since the pool is just being used to execute subprocesses, you're probably better off using a thread pool, but up to you. Otherwise looks good to me

PR #252, opened 2023-01-10
Vt tools

EthanMarx:
- Adds tools for calculating VT integrals for general distributions in the `analysis` library
- Adds VT vs FAR plot to the `vizapp`

- [ ] Find a better home for `priors.py`


libs/analysis/bbhnet/analysis/vt.py
```diff
@@ -0,0 +1,163 @@
...
+        # if no target distribution is passed,
+        # use the source distribution
+        if target is None:
+            target = self.source
+
+        weights = []
+        for sample in self.recovered_parameters:
+            # calculate the weight for each sample
+            # using the source and target distributions
+            weight = target.prob(sample) / self.source.prob(sample)
```
deepchatterjeeligo, line 136: I think this might be fine. My concern was the dimensions not matching. But just did a quick check.

```
prior_a = bilby.prior.Uniform(1, 101)
prior_b = bilby.prior.Uniform(1, 101)
prior_c = bilby.prior.Uniform(0, 0.5)

source = bilby.prior.PriorDict(dict(a=prior_a, b=prior_b))
target = bilby.prior.PriorDict(dict(a=prior_a, b=prior_b, c=prior_c))

>>> source.prob(dict(a=5, b=6))
0.0001
>>> target.prob(dict(a=5, b=6))
0.0001
```
It looks like if the target prior has more number of dimensions that `recovered_parameters` it will marginalize over those.

However, to avoid confusion/assumptions, it might be better to pass only the `weighting_parameter_keys`, which is assumed to be present in both
```
source_subset = PriorDict({k: self.source[k] for k in weighting_parameter_keys})
target_subset = PriorDict({k: target[k] for k in weighting_parameter_keys})

weight = target_subset(sample) / source_subset(sample)
```
```

Comment (EthanMarx) [reply libs/analysis/bbhnet/analysis/vt.py:136]: Awesome thanks for looking into this. Makes sense

File: libs/analysis/bbhnet/analysis/vt.py [hunk 2]
```diff
@@ -0,0 +1,166 @@
...
+    theta_max, theta_min = (
+        np.pi / 2 - dec_min.value,
+        np.pi / 2 - dec_max.value,
+    )
+    omega = -2 * math.pi * (np.cos(theta_max) - np.cos(theta_min))
+
+    # calculate the volume of the universe
+    # over which injections have been made
+    integrand = (
+        lambda z: 1.0 * (cosmology.differential_comoving_volume(z)).value
```
deepchatterjeeligo, line 59: We had a brief chat about this. Turns out the `UniformInComovingVolume` in bilby does not have the time dilation factor. What they call `UniformInSourceFrame` is what is uniform in comoving volume typically in GW literature. See: https://git.ligo.org/lscsoft/bilby/-/blob/master/bilby/gw/prior.py#L311-346

Hence, let's keep the prior for distance as uniform in source frame, but use the 1/(1 + z) factor here.

```suggestion
        lambda z: 1 / (1 + z) * (cosmology.differential_comoving_volume(z)).value
```

projects/sandbox/vizapp/pyproject.toml
```diff
@@ -19,6 +19,10 @@ typeo = {git = "https://github.com/ML4GW/typeo.git", branch = "main"}
 
 "bbhnet.analysis" = {path = "../../../libs/analysis", develop = true}
 "bbhnet.logging" = {path = "../../../libs/logging", develop = true}
+bilby = "^1.4.1"
+
+[tool.poetry.group.dev.dependencies]
```
alecgunny, line 24: Do we want to keep this dep or was this just for that notebook that's leftover?

libs/analysis/bbhnet/analysis/vt.py
```diff
@@ -0,0 +1,166 @@
...
+
+        Returns tuple of (vt, std, n_eff)
+        """
+        weights = self.weights(target)
+
+        mu = np.sum(weights) / self.n_injections
+
+        v0 = self.livetime * YEARS_PER_SECOND * self.volume
+        vt = mu * v0
+        variance = v0**2 * (
```
alecgunny, line 160: I might suggest breaking this up into a couple lines to help make it more clear but not a big deal

projects/sandbox/vizapp/vizapp/plots/vt.py
```diff
@@ -0,0 +1,179 @@
...
+
+        self.fars = np.logspace(0, 7, 7)
+        self.configure_sources()
+        self.configure_widgets()
+        self.configure_plots()
+        self.logger = logging.getLogger("VT vs FAR")
+        self.logger.debug(self.fars)
+
+    def configure_plots(
+        self,
```
alecgunny, line 43: really minor but if you get rid of the comma black will let you keep this all on the same line

alecgunny (changes requested):
Looks like there's a leftover debugging notebook in there? Otherwise just a couple aesthetic comments but not blockers

projects/sandbox/vizapp/pyproject.toml
EthanMarx: Actually wanted to ask about this. I added the `priors.py` from the `datagen` project into the vizapp project so we don't have to carry the `datagen` dep (which includes a `torch` dep). I guess the options are to make the `torch` dep an `extra` in the datagen project, or to refactor out `priors.py`.

alecgunny: I think refactoring out `priors` definitely makes a lot of sense, especially since bilby can be managed by poetry. It an just be a one-file `bbhnet` library

EthanMarx: Okay. Maybe this should be done in #243 @wbenoit26

wbenoit26: If #243 looks good as-is, I'd rather get that merged and then do a quick second PR to refactor the priors

alecgunny: I would agree that probably makes sense

alecgunny:
Got the app to work for me, but note there's no `source_prior` specified in `projects/sandbox/pyproject.toml` which caused an error at first

EthanMarx:
A couple important things

1. As Will discovered, using the `BBHPriorDict` automatically enforces m1 > m2 which is great. 
2. It's important to remember that mass ratio is defined as `q = m2 / m1`, so that q <= 1. This `Constraint` in the Bilby priors should always be less than 1. I changed the mass ratio `Constraint` from( 0.2, 5) to (0,1) since the vizapp was breaking due to injections outside the constraint range. 

Given all of these new developments, we should probably regenerate our timeslide data using the `end_o3_ratesandpops` prior.

wbenoit26:
The `BBHPriorDict` doesn't enforce m2 < m1 by itself, it still needs a `Constraint`. It's just that if you put a constraint on the mass ratio, the `BBHPriorDict` will know how to interpret that to give you an appropriate pair of masses.

Also, a more minor note: I switched to a prior over redshift instead of luminosity distance, as sampling a `BBHPriorDict` will automatically generate the latter given the former, but not the other way around.

EthanMarx:
Ah okay got it that makes sense. 

There are a few other things i want to add, but let's just get this in, and I'll work on them in a separate PR. I was able to get it up on my machine but If someone wants to run it that would be great.

alecgunny (approved):

PR #253, opened 2023-01-12
constraining scipy version due to astropy issue

alecgunny:
As I note in the `pyproject.toml` of `projects/sandbox/datagen`, astropy conflicts with `scipy>=1.10` due to astropy/astropy#14232 . Once this gets put in a release and we can start pinning to that version, we can remove this constraint, but for now we need it.

### DIFF
--- a/libs/data/tests/test_waveform_injection.py
+++ b/libs/data/tests/test_waveform_injection.py
@@ -5,7 +5,7 @@ import torch
 from bbhnet.data.waveform_injection import BBHNetWaveformInjection
 
 
-def sample(obj, N, _):
+def sample(obj, N, dev=None):
     return torch.ones((N, 2, 128 * 2)), None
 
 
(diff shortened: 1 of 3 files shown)


EthanMarx (approved):

deepchatterjeeligo:
Saw this notification on slack. We had the same issue in a different project with regard to astropy cosmology. This has been fixed in astropy 5.2.1

alecgunny:
@deepchatterjeeligo awesome good to know, looks like that's available on PyPI now too. Might be worth opening an issue to upgrade this and confirm that it doesn't break with any of our other dependencies so that we can relax these constraints again

deepchatterjeeligo:
sure. Are we using astropy. I don't see it being used in datagen. Or is the astropy dependency coming through bilby?

alecgunny:
bilby and gwpy I believe

PR #260, opened 2023-01-23
update default priors

EthanMarx:
- Hard codes `Planck15` cosmology for all of our priors. (Should think about best way to add this to the config)
- Adds `spin_bbh` prior that includes the spin priors from `end_o3_ratesandpops` (i.e allows for non zero spins)
- Makes `spin_bbh` the default for generating training waveforms
- Makes `end_o3_ratesandpops` the default for generating timeslide injections

### DIFF
--- a/libs/priors/bbhnet/priors/priors.py
+++ b/libs/priors/bbhnet/priors/priors.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 from typing import Sequence, Tuple
 
+import astropy.cosmology as cosmo
 import h5py
 import numpy as np
 from bilby.core.prior import (
@@ -12,11 +13,7 @@ from bilby.core.prior import (
     Sine,
     Uniform,
 )
-from bilby.gw.prior import (
-    BBHPriorDict,
-    UniformComovingVolume,
-    UniformSourceFrame,
-)
+from bilby.gw.prior import BBHPriorDict, UniformSourceFrame
 
 # Unit names
 msun = r"$M_{\odot}$"
@@ -108,7 +105,9 @@ def nonspin_bbh() -> BBHPriorDict:
     prior["mass_1"] = Uniform(5, 100, unit=msun)
     prior["mass_2"] = Uniform(5, 100, unit=msun)
     prior["mass_ratio"] = Constraint(0, 1)
-    prior["redshift"] = UniformSourceFrame(0, 0.5, unit=mpc, name="redshift")
+    prior["redshift"] = UniformSourceFrame(
+        0, 0.5, unit=mpc, name="redshift", cosmology=cosmo.Planck15
+    )
     prior["psi"] = 0
     prior["a_1"] = 0
     prior["a_2"] = 0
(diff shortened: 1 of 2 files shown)


EthanMarx:
@deepchatterjeeligo @wbenoit26 Thoughts on mass ratio lower limits? What do the CBC template banks use?

libs/priors/bbhnet/priors/priors.py
```diff
@@ -120,12 +119,33 @@ def nonspin_bbh() -> BBHPriorDict:
...
+    return prior
+
+
 def end_o3_ratesandpops() -> BBHPriorDict:
     prior = uniform_extrinsic()
     prior["mass_1"] = PowerLaw(alpha=-2.35, minimum=2, maximum=100, unit=msun)
     prior["mass_2"] = PowerLaw(alpha=1, minimum=2, maximum=100, unit=msun)
     prior["mass_ratio"] = Constraint(0.02, 1)
-    prior["redshift"] = UniformComovingVolume(0, 2, unit=mpc, name="redshift")
+    prior["redshift"] = UniformSourceFrame(
```
wbenoit26, line 146: Why did you change the prior here? It looks like the prior used in the rates and pops analysis was the comoving volume

projects/sandbox/pyproject.toml
```diff
@@ -79,7 +79,7 @@ reference_frequency = "${base.reference_frequency}"
 datadir = "${base.datadir}"
 logdir = "${base.logdir}"
 n_samples = 10000
```
wbenoit26, line 81: Could you swap this to `num_signals`? I changed the variable name in the project but forgot to update it here.

wbenoit26 (changes requested):

wbenoit26:
I think I usually see lower mass ratio limits of about 0.02

libs/priors/bbhnet/priors/priors.py
EthanMarx: @deepchatterjeeligo has mentioned that Bilby's `UniformComovingVolume` is different than what is traditionally considered "UniformComovingVolume" (which seems weird). Can you link to where you've seen that prior implemented

wbenoit26: I'm basing this off what they describe [here](https://wiki.ligo.org/CBC/RatesPop/EndO3PopulationInjections), under the end o3 part of the BBH 'Broad' Injections. They say that p(z)~dV_c/dz * dt_source/dt * (1+z)^k. They chose k = 1, this it simplifies to p(z)~dV_c/dz, which is how bilby defines their uniform comoving volume prior. I'll try to find their implementation of it

wbenoit26: [This](https://git.ligo.org/RatesAndPopulations/lvc-rates-and-pop/-/blob/master/bin/endo3_distributions.py#L164) is where they implement their prior on z, but I'm having trouble finding documentation for that lal function.

PR #261, opened 2023-01-25
Fixing train/validation/test disparity issues

alecgunny:
Resolving issues with the disparity between training, validation, and test performance by switching to GroupNorm normalization and increasing the amount of training/validation background data. Other notable changes are:
- Un-swapping plus and cross polarizations during training, which were being accidentally switched
- Increasing NN layers to `[2, 2, 2, 2]` to reflect ResNet18 architecture
- Switching to latest versions of TensorRT/Triton to avoid server deadlock issues, but switching to full precision inference due to underflow issues in latest version of TensorRT
- Moving to a ramp-then-decay learning rate scheduler, which is more consistent with the literature especially for larger batch sizes

Remaining tasks
- [x] Update `export` lockfile to reflect new TensorRT version
- [ ] Implement validation procedure similar to the one done [here](https://github.com/alecgunny/ml4gw-examples/blob/main/training.ipynb)
- [x] Run the pipeline end-to-end to confirm everything runs


wbenoit26:
Worth noting that the larger signal dataset (and maybe the background datasets) will likely run into memory issues if they're just done on one of the nodes

alecgunny:
@wbenoit26 been running this afternoon on the Hanford DGX node and haven't hit any issues

alecgunny:
(well some issues with glitch generation but unrelated to memory load)

wbenoit26:
The signal file should be about 12-13GB, and that's all in memory at once, right? We were seeing issues with the timeslides at around 6-7GB, for that ~30000 second segment (which is why I was maybe worried about the background size). It might be fine, but I think it could be an issue depending on cluster usage

alecgunny:
Ah true ok that makes sense, and FWIW that ~30000s segment (34628s to be exact) is now the training segment (32768s of it anyway). Out-of-memory dataloaders have been in the back of my mind for a while now, so if this breaks for folks on other clusters maybe it's time to prioritize that.

alecgunny:
This pipeline is able to run basically end-to-end. The only hang up is on glitch generation, which is still experiencing some issues that I haven't been fully able to flush out due to logging issues (see ML4GW/mldatafind#30). I'll re-run now with the logging fixed to get a better sense of what's going wrong, but since that's out of scope for this PR I think this should be basically good to go.

I still haven't implemented the validation from my repro notebook, but the current validation methods produce good models consistent with the scores seen from the notebook model, so I'm prepared to say this is working and filing an issue to improve validation with actual loss measurement and possibly AUC (if it does a better job of selection than just recall at FAR=0).

libs/trainer/bbhnet/trainer/trainer.py
```diff
@@ -180,10 +180,17 @@ def train(
     # TODO: Allow different loss functions or optimizers to be passed?
     criterion = torch.nn.functional.binary_cross_entropy_with_logits
     optimizer = torch.optim.Adam(
-        model.parameters(), lr=lr, weight_decay=weight_decay
+        model.parameters(), lr=max_lr, weight_decay=weight_decay
```
EthanMarx, line 183: I presume the `lr` set here is discarded, (and not used as the starting `lr`), but wanted to clarify.

projects/sandbox/pyproject.toml
```diff
@@ -16,14 +17,14 @@ datadir = "${DATA_DIR}"
...
-test_stop = 1262751694  
+glitch_start = 1262653854
+train_start = 1262653854
+train_stop = 1262686622 # 32768 s segment
+test_stop = 1262945822 # 3 day window for test
 channel = "DCS-CALIB_STRAIN_CLEAN_C01"
 frame_type = "HOFT_C01"
 state_flag = "DCS-ANALYSIS_READY_C01:1"
-resnet = {layers = [3, 3, 3]}
+resnet = {layers = [2, 2, 2, 2], norm_groups = 8}
```
EthanMarx, line 27: My guess is this is set to `> n_ifos` to account for future layers where more channels are introduced?

EthanMarx (approved):
This should be merged regardless, but have you found a resolution to the odd behavior seen in the vizapp?

libs/trainer/bbhnet/trainer/trainer.py
alecgunny: Yeah my understanding is the scheduler just disregards whatever the optimizers existing LR is and sets it based on its schedule (in this case it gets initialized to `max_lr / div_factor`, where `div_factor` defaults to 25)

projects/sandbox/pyproject.toml
alecgunny: So the norm layer I've built will use `min(in_channels, norm_groups)`, so for the first layer this will equal to `num_ifos`, and then `norm_groups` after that. The number of groups to use for the later layers is a hyperparameter that will need to be tuned, this is just a random first guess that seems to work decently well.

alecgunny:
Haven't had a chance to dig through the data yet, but with the current SNR distribution this seems to be working so I'll merge but will file an issue for that and for the validation.

PR #268, opened 2023-01-31
Deprecated data lib

### LINKED ISSUE #262: Update `ml4gw` and deprecate `bbhnet.data` library
Now that `ml4gw` has the whitening transform implemented, we should use it directly from there and get rid of the `bbhnet.data` library. All the light subclasses of `ml4gw` modules can be implemented in the `train` project. Other tasks include
- Updating to the latest `.fit` syntax for both the whitening and injection modules
- Making sure `bbhnet.architectures` points to `ml4gw` for the whitening transform as well

wbenoit26:
Closes #262


projects/sandbox/train/train/data_structures.py
```diff
@@ -71,3 +74,65 @@ def __next__(self) -> Tuple[torch.Tensor, torch.Tensor]:
...
+        if not self.training:
+            return X, y
+
+        X, idx, _ = super().forward(X)
+        y[idx] = 1
+        return X, y
+
+
+# TODO: generalize to arbitrary ifos
+class GlitchSampler(Transform):
```
alecgunny, line 90: I think you can get rid of the `transforms` module and have this inherit directly from `torch.nn.Module`. Below for the parameters I would switch from `**glitches` to `*glitches` since we don't use the IFO name information and this makes it easier to make sure things are passed in the correct order (e.g. a user couldn't actually use the `**` syntax when initializing since they can't guarantee the dict they're using will maintain the order they want). Then for registering the params you could something like
```python
super().__init__()
for i, glitch in enumerate(glitches):
    glitch = torch.Tensor(glitch)
    self.register_buffer(f"glitch_{i}", glitch)
```
Then during `forward` you can do
```python
for i, glitch in enumerate(self.buffers()):
    ...
```

The reality is that we should probably be treating the glitches as `Tensor`s and keeping them on CPU then moving the sampled glitches to GPU before insertion, not `Buffer`s or `Parameter`s which get moved to whatever device the layer is on, but our glitch datatsets are small enough right now that this shouldn't be that big of a deal.

alecgunny: You can also get rid of the `TODO` since this is now fully general to any number of IFOs

alecgunny: Actually now that I'm seeing how this actually gets initialized below, I'm realizing that this would require specifying everything as args, e.g. `GlitchSampler(glitch_prob, max_offset, h1_glitches, l1_glitches)`, which is maybe less clear than allowing the user to specify `prob=..., max_offset=...`, so maybe just keep this for now and we can figure out a better way of doing this down the line.

Either way I would say drop the `transform` module and just switch the loop above to be
```python
for ifo, glitch in glitches.items():
    self.register_buffer(ifo, torch.Tensor(glitch))
```

alecgunny (commented):
See note about dropping `transform` module, otherwise looks great

wbenoit26:
@alecgunny Running into some issues because the `Whitening` module needs to be fit before it's called, and I'm not sure where that should be happening

alecgunny:
@wbenoit26 if you don't care about what value the time domain filter (and by extension the output of layer) has, you can get around the fitting step by just calling
`whitener.build(time_domain_filter=whatever_value_you_want_to_set_the_time_domain_filter_to)`

wbenoit26:
Unless I'm doing this wrong, I don't think that will work. `forward` requires that `kernel_length` be determined, and that happens during fitting.

To be explicit, the issue is coming from (I think) `hermes` trying to determine the output shape of the model

alecgunny:
Oh that's right sorry, you can set that explicitly as well
```python
whitener.build(
    time_domain_filter=whatever_value_you_want_to_set_the_time_domain_filter_to,
    kernel_length=torch.Tensor((whatever_the_kernel_length_is,))
)
```
Alternatively, you can just `fit` the whitener on some dummy background data before you run `forward`. FWIW, `hermes` infers the output shape by just running a tensor of the specified input shape through the model, so all this tells you is that the forward step of your layer isn't working.

wbenoit26:
No, I tried explicitly setting the kernel length and got the same error. Here, I've pushed a commit with the `build` step in so you can see what I'm seeing. It seems like the tdf needs to be a specific length, and that the kernel length that I'm passing isn't affecting the `forward` step.

I also tried fitting with dummy data, but still got the same error. How does `self.kernel_length` get set in `fit`? Not familiar with `buffer`s, so I'm not sure how that works.

alecgunny:
Ohhhh ok got it, I didn't realize this was all during actual export I figured this was during testing, sorry about that. So for this the preprocessor _should_ be getting "built" when you load in the weights, i.e. two lines later when you call `nn.load_state_dict`, but evidently this isn't happening. I'll pull your fork and take a more in-depth look

alecgunny:
Actually now I'm realizing this is an `ml4gw` issue. I guess `load_state_dict` doesn't get called when you're a submodule of a larger module, which is how I was making sure `self.built` got set to `True`. I should be able to turn this around pretty quick

alecgunny:
This will be fixed with ML4GW/ml4gw#42

alecgunny:
Ok so the `built` issue is fixed, you just need to make sure to set the `kernel_length` parameter to the right value, so should be sufficient to go in `get_network_weights`
```python
preprocessor.whitener.build(kernel_length=torch.Tensor((kernel_length,)))
```

wbenoit26:
This raises some new errors, maybe related to padding? I'm trying to parse what's going on

alecgunny (approved):

PR #270, opened 2023-02-06
adding README and updating contributing instructions

alecgunny:
I think this is just missing a description of the evaluation process but don't know how much detail is necessary to get into at the root level, so maybe leave this for the future as we work out the documentation for the projects.


README.md
```diff
@@ -0,0 +1,169 @@
+# BBHNet
+Detecting binary blackhole mergers from gravitational wave strain timeseries data using neural networks, with an emphasis
```
wbenoit26, line 2: emphasis on*

wbenoit26: Also, throughout, black hole should be two words

wbenoit26 (commented):
I found only a couple of typos, so I'm good with this once the tests are passing

alecgunny:
@wbenoit26 for some reason the pre-commit hook isn't showing as finished here, even though it shows as passing under the Actions tab. The unit-tests tab shows a failure there, but that's only because the lib matrix is empty because there's no library changes to test. I would suggest therefore that this is ready to go from a testing standpoint, even if we have to technically bypass the branch protections to merge it.

wbenoit26 (approved):
Sounds good to me

PR #272, opened 2023-02-07
Sensitive calculation improvements

EthanMarx:
1. Parameterizes and standardizes the `cosmology` used across the whole analysis.
2. New `PriorDict` class. The user can declare if the prior was defined in the source or detector frame. When sampling, the user must pass the frame to sample in. Whether or not the conversion is necessary is handled under the hood. 
3. VT has been changed to just a sensitive volume calc. The VT can be reimplimented trivially by multiplying by analyzed livetime.

### DIFF
--- a/libs/priors/bbhnet/priors/priors.py
+++ b/libs/priors/bbhnet/priors/priors.py
@@ -32,7 +32,7 @@ class PriorDict(BBHPriorDict):
     def sample(self, n: Optional[int] = None, source_frame: bool = False):
         samples = super().sample(n)
         # only rescale if requested frame is different from defined frame
-        rescale = self._source_frame != source_frame
+        rescale = self._source_frame or source_frame
         if not rescale:
             return samples
 
@@ -43,10 +43,20 @@ class PriorDict(BBHPriorDict):
         samples["mass_2"] *= factor
         return samples
 
-    # TODO: implement prob method that accounts for jacobian
-    # which will be used during importance sampling
-    def prob(self):
-        raise NotImplementedError
+    def prob(self, samples, source_frame: bool = False):
+        """
+        Override the default prob method to account for the jacobian
+        between detector frame and source frame masses.
+        See equation 7) in https://arxiv.org/pdf/1712.00482.pdf
+        """
+        prob = super().prob(samples)
+        rescale = self._source_frame or source_frame
+        if not rescale:
+            return prob
+
+        factor = (1 + samples["redshift"]) ** 2
+        jacobian = factor if source_frame else (1 / factor)
+        return prob * jacobian
 
 
 def uniform_extrinsic(source_frame: bool = False) -> PriorDict:
--- a/libs/priors/bbhnet/priors/priors.py
+++ b/libs/priors/bbhnet/priors/priors.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Sequence, Tuple
+from typing import TYPE_CHECKING, Sequence, Tuple
 
 import h5py
 import numpy as np
@@ -13,11 +13,10 @@ from bilby.core.prior import (
     Sine,
     Uniform,
 )
-from bilby.gw.prior import (
-    BBHPriorDict,
-    UniformComovingVolume,
-    UniformSourceFrame,
-)
+from bilby.gw.prior import BBHPriorDict, UniformSourceFrame
+
+if TYPE_CHECKING:
+    from astropy.cosmology import Cosmology
 
 # Unit names
 msun = r"$M_{\odot}$"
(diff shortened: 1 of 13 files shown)


EthanMarx:
- [ ] Tests for `prob` calculation that accounts for jacobian

projects/sandbox/datagen/datagen/scripts/timeslides.py
```diff
@@ -305,6 +310,19 @@ def main(
                 future = submit_write(pool, injection_ts, t, **injected_data)
                 futures.append(future)
 
+                # infer redshift based on injected luminosity distance and
+                # cosmology use redshift to calculate source frame masses
+                parameters["redshift"] = cosmo.z_at_value(
```
wbenoit26, line 315: Do we need this? Looks like all the priors specify things in terms of redshift anyway, so `parameters["luminosity_distance"]` is coming from the inverse of this calculation

libs/priors/bbhnet/priors/priors.py
```diff
@@ -1,101 +1,66 @@
...
     Uniform,
 )
-from bilby.gw.prior import (
-    BBHPriorDict,
-    UniformComovingVolume,
-    UniformSourceFrame,
-)
+from bilby.gw.prior import BBHPriorDict, UniformSourceFrame
+
+from bbhnet.priors.utils import read_priors_from_file
```
wbenoit26, line 15: This file, and probably a test_utils file, need to be committed

wbenoit26 (commented):
I know you're not quite done with this, but adding some comments now

projects/sandbox/datagen/datagen/scripts/timeslides.py
EthanMarx: Yeah this should be flipped. I think it's safe to assume that our testing set priors will be defined with redshift as opposed to luminosity distance.

libs/priors/bbhnet/priors/priors.py
```diff
@@ -1,101 +1,66 @@
...
+        between detector frame and source frame masses.
+        See equation 7) in https://arxiv.org/pdf/1712.00482.pdf
+        """
+        prob = super().prob(samples)
+        rescale = self._source_frame or source_frame
+        if not rescale:
+            return prob
+
+        factor = (1 + samples["redshift"]) ** 2
+        jacobian = factor if source_frame else (1 / factor)
```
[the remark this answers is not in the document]
EthanMarx: Since we define our priors with redshift I don't think we need that factor. 
If we were to have defined our prior in luminosity distance and then converted to redshift, that factor would matter.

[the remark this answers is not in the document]
wbenoit26: Ah, gotcha, that makes sense

projects/sandbox/datagen/datagen/scripts/timeslides.py
```diff
@@ -305,6 +310,19 @@ def main(
                 future = submit_write(pool, injection_ts, t, **injected_data)
                 futures.append(future)
 
+                # infer luminosity distance based on redshift and cosmology
+                parameters["luminosity_distance"] = cosmo.luminosity_distance(
```
wbenoit26, line 314: As long as we're always inheriting from a cosmologically-aware `bilby` prior, we can skip this step - `bilby` will calculate `luminosity_distance` from `redshift` if the former doesn't exist. On the other hand, this enforces that our desired cosmology is used. `bilby` will use the cosmology specified in the `redshift` prior, but if that's ever not set, it will default to Planck 15 (which is currently what you've set, but in case that ever changes, this would add backup)

projects/sandbox/pyproject.toml
```diff
@@ -174,7 +175,7 @@ min_segment_length = 1024
 
 # timeslide parameters
 shifts = [0, 1]
-n_slides = 10
+n_slides = 30
```
wbenoit26, line 178: Was this an intended change?

projects/sandbox/datagen/datagen/scripts/timeslides.py
EthanMarx: Ah okay. So will the `luminosity_distance` field automatically be populated?

wbenoit26: Yup, during sampling

wbenoit26: Correcting myself here as well as on Slack for completeness: the behavior I described doesn't happen. From the documentation, I believe that it's supposed to, but it does not. This conversion should stay.

EthanMarx:
As usual, bit off more than I could chew. I removed the generalization I made to the priors class as I found itcreated more confusion than it removed. 

This PR now only does the following:
1. VT is changed to just Sensitive Volume
2. Adds a `cosmology` parameter to the `pyproject.toml` that get's passed at the appropriate places in the pipeline.

projects/sandbox/pyproject.toml
```diff
@@ -233,5 +236,6 @@ train_data_dir = "${base.datadir}"
 sample_rate = "${base.sample_rate}"
 fduration = "${base.fduration}"
 valid_frac = "${base.valid_frac}"
-logdir = "${base.basedir}"
+logdir = "./"#"${base.basedir}"
```
wbenoit26, line 239: I think this change was left in from testing

PR #282, opened 2023-02-07
Minor glitch script bug fixes

EthanMarx:
- Fixes how glitches are appended when querying strain
- Only queries glitches that are within half window of chunk edges


alecgunny (approved):
Looks good to me, what was this issue this is addressing?

EthanMarx:
Two things that got mixed up after the `mldatafind incorporation`

- We were appending the `TimeSeries` objects as opposed to arrays
- Can only select trigger times a half window inside each chunk edge to ensure we have enough data.

PR #284, opened 2023-02-09
Fixed glitch sampler bug

wbenoit26:
Quick PR to fix a bug I introduced when getting rid of the `data` library. I changed how the `GlitchSampler` class stored its glitches, but didn't update `forward`


alecgunny (commented):
Looks great, good catch. Is there any way we can copy over some of the old glitch sampler tests to make sure this stays working?

wbenoit26:
Holy cow I really just deleted those tests and didn't move them over. Wow. Yeah, I'll add them back in.

projects/sandbox/train/tests/test_data_structures.py
```diff
@@ -0,0 +1,187 @@
...
+        if prob == 0:
+            assert (inserted == 0).all().item()
+        elif prob == 1:
+            assert (inserted != 0).all().item()
+        else:
+            assert (inserted[:2] != 0).all().item()
+            assert (inserted[2:] == 0).all().item()
+
+
+def sample(obj, N, dev=None):
```
alecgunny, line 159: the `sample` method actually got rid of the `dev` kwarg so we can actually just drop this. FWIW `sampled_params` would have shape `(N, 3)` but since we don't actually use it for anything this doesn't really matter either.

alecgunny (commented):
Looking at this now, we don't use either the base `Transform` class or `StandardScaler` (which is implemented in `ml4gw` anyway) from the `transforms` submodule do we? Probably makes sense to just ditch this and the associated tests

projects/sandbox/train/tests/test_data_structures.py
wbenoit26: Sounds good, updated

wbenoit26:
You mean get rid of everything in that `transforms` directory? Yeah, I was going to raise that question - I think it makes sense to.

alecgunny (approved):

PR #286, opened 2023-02-10
Move away from bbhpriordicts

wbenoit26:
I've swapped the `BBHPriorDict`s in our priors back to ordinary `PriorDict`s, and added a conversion function to enforce a mass ratio constraint. This will prevent extraneous parameters from being set by the default prior dictionaries.

I also removed the test on `priors.py`, as they were left over from #272 when Ethan had created a custom prior class.

Here are a few plots that demonstrate that these priors have the intended behavior. Let me know if anyone would like to see others.
image
image
image


libs/priors/bbhnet/priors/utils.py
```diff
@@ -3,8 +3,15 @@
...
-from bilby.core.prior import Interped
-from bilby.gw.prior import BBHPriorDict
+from bilby.core.prior import Interped, PriorDict
+
+
+def mass_ratio_constraint(samples):
+    if "mass_1" not in samples.keys() or "mass_2" not in samples.keys():
+        raise KeyError("mass_1 and mass_1 must exist to have a mass_ratio")
+    out_samples = samples
+    out_samples["mass_ratio"] = samples["mass_2"] / samples["mass_1"]
```
EthanMarx, line 13: Is there a reason for reassigning variable here? If so, I think you actually wan't `out_samples = samples.copy()`, what you've done here is basically add another reference to the dictionary data structure samples was pointing to, so altering `out_samples` will also alter `samples`

EthanMarx:
Thanks for doing this. The last thing I'll request is to make the same scatter plot for the detector frame masses of `end_o3ratesandpop`. So take the plot you have, and multiple each mass by the sampled (1 + z). We should then use this to define our training prior

libs/priors/bbhnet/priors/utils.py
wbenoit26: Yeah, you're correct, should be `samples.copy()`

wbenoit26:
@EthanMarx Here's that plot:
image

libs/priors/bbhnet/priors/utils.py
alecgunny: Yeah if this is all happening internally to bilby we don't necessarily need to a do a copy here do we? Obviously not super important but just curious

wbenoit26: It ends up not mattering because `bilby`'s conversion function doesn't generate new parameters the way that  (I think) it's supposed to.

wbenoit26:
@EthanMarx Is this good to go?

EthanMarx:
Instead of having `detector_frame_prior` be an argument to many of these functions, can we have the `prior` Callables also return this boolean? e.g. 
```
prior, detector_frame_prior = prior()
```
This way, detector_frame_prior is defined right alongside the prior itself, and we don't have to remember which prior is what.

wbenoit26:
Sure, that should be taken care of now.

EthanMarx (approved):

PR #287, opened 2023-02-10
Rejected hopeless injections

EthanMarx:
Rejects hopeless injections based on estimate of network SNR from hplus.

### DIFF
--- a/projects/sandbox/datagen/datagen/scripts/timeslides.py
+++ b/projects/sandbox/datagen/datagen/scripts/timeslides.py
@@ -29,6 +29,7 @@ from bbhnet.io.timeslides import TimeSlide
 from bbhnet.logging import configure_logging
 from bbhnet.parallelize import AsyncExecutor
 from ml4gw.gw import compute_ifo_snr, compute_observed_strain, get_ifo_geometry
+from ml4gw.spectral import normalize_psd
 
 
 @scriptify
@@ -56,6 +57,7 @@ def main(
     reference_frequency: float = 20,
     waveform_approximant: str = "IMRPhenomPv2",
     fftlength: float = 2,
+    hopeless_snr_threshold: float = 0.0,
     state_flag: Optional[str] = None,
     force_generation: bool = False,
     verbose: bool = False,
@@ -186,9 +188,6 @@ def main(
                 buffer_,
                 spacing,
             )
-            waveform_it = waveform_iterator(
-                pool, sampler, waveform_generator, num_shifts
-            )
 
             # begin the download of data in a separate thread
             logging.debug(f"Beginning download of segment {seg_str}")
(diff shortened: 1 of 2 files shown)


projects/sandbox/datagen/datagen/utils/timeslides.py
```diff
@@ -47,6 +50,11 @@ def __call__(self):
         times = self.signal_times + jitter
 
         params = self.prior.sample(self.num_signals)
+
+        # our testing prior is defined in the source frame, so we
+        # have to convert to detector frame masses.
+        params["mass_1"] = params["mass_1"] * (1 + params["redshift"])
+        params["mass_2"] = params["mass_2"] * (1 + params["redshift"])
```
wbenoit26, line 57: This will conflict with the way I've done the conversion in #286

projects/sandbox/datagen/datagen/utils/timeslides.py
```diff
@@ -70,23 +78,62 @@ def __call__(self, parameters):
...
+        params = sampler()
+        signals = generator(params)
+
+        # crude estimate of snrs using hplus
+        snrs = compute_network_snr(
+            signals[:, 0.0:], psds, sample_rate, highpass
+        )
+        signals = signals[snrs > hopeless_snr_threshold]
+        n_rejected += np.sum([snrs < hopeless_snr_threshold])
+        waveforms.append(signals)
```
wbenoit26, line 101: We could end up with a lot more signals than can fit in the segment with the desired spacing. We'll also need a way to make sure that signals don't overlap.

EthanMarx: All of the logic that determines the number of signals based on the segment length and spacing is already controlled by the `Sampler`

wbenoit26: I'm missing something - how does that work? If you reject, say, 25% of the signals generated each time, won't `len(waveforms)` be `1.5 * num_signals`?

EthanMarx: I see the confusion. This commit doesnt have the `waveforms = waveforms[:n_signals]` thats on my local repo.

wbenoit26: Ah okay, I see

EthanMarx:
Resolved memory issue here by changing the minimum spacing from 0 to 30. Not ideal for robustness but given this will all be reworked very shortly I think we just review this as is.

projects/sandbox/pyproject.toml
```diff
@@ -1,10 +1,10 @@
...
-    "datagen:generate-background",
-    "datagen:generate-waveforms",
-    "datagen:generate-glitches",
-    "train:train:resnet",
-    "export:export-model:resnet",
+    #"datagen:generate-background",
+    #"datagen:generate-waveforms",
+    #"datagen:generate-glitches",
+    #"train:train:resnet",
+    #"export:export-model:resnet",
```
wbenoit26, line 7: These can be uncommented

projects/sandbox/pyproject.toml
```diff
@@ -187,7 +187,7 @@ spacing = 48
 buffer = 4
 jitter = 2
 cosmology = "${base.cosmology}"
-
+hopeless_snr_threshold = 6
```
wbenoit26, line 190: Have we confirmed that this is the typical threshold? Deep was saying he thought it might be 4. Not a big hangup, but worth checking if we haven't already.

projects/sandbox/datagen/datagen/scripts/timeslides.py
```diff
@@ -29,6 +30,10 @@
 from bbhnet.logging import configure_logging
 from bbhnet.parallelize import AsyncExecutor
 from ml4gw.gw import compute_ifo_snr, compute_observed_strain, get_ifo_geometry
+from ml4gw.spectral import normalize_psd
+
+# suppress annoying bilby log when calling .sample
+setup_logger(log_level="WARNING")
```
wbenoit26, line 36: Love it

projects/sandbox/datagen/datagen/utils/timeslides.py
wbenoit26: This conversion is already taken care of in `injection.py`

wbenoit26 (changes requested):
Other than the mass conversion, this looks good to me. @deepchatterjeeligo had some good ideas in our today about how we can speed up the process of generating non-hopeless (hopeful?) waveforms. Not sure how difficult they would be to implement, so we could get this PR in first, and then add improvements in a second PR, or it might make sense to add them here.

alecgunny (commented):
Looks good to me but @wbenoit26 I think will have better insight into the actual GW stuff. Did we figure out what was causing the test crashes?

EthanMarx:
It was a memory problem, although I never tracked the exact source. I resolved by changing the minimum spacing between injections from 0 to 30 in the tests.

Although we have to generate more waveforms _in total_, there shouldn't have been that many more waveforms in memory _at once_ than the previous implementation.

wbenoit26 (approved):

Issue #290, opened 2023-02-14
`vizapp` and sensitive volume bugs


alecgunny:
In the `vizapp`:
- `app.App` needs to save `self.cosmology` in `__init__`
- In [these lines](https://github.com/ML4GW/BBHNet/blob/2dd3baac0759552be5fe20de832a4a3cbcbfe86f/projects/sandbox/vizapp/vizapp/plots/vt.py#L140-L144) in `vt.py`, the mass fields should refer to `self.foreground.m1s` and `self.foreground.m2s`. The `Foreground` object also doesn't record a redshift at the moment. This can either be added as a field and calculated at initialization, or we can just do the conversion from luminosity distance to redshift here

The other major issue is that it seems like the first initialization of `SensitiveVolumeCalculation` takes ~25 minutes or so, likely on the `v0` calculation. Not sure why this is, but seems like we should just be initializing this once and then calling it with different sets of recovered parameters. Something like

```python
@dataclass
class SensitiveVolumeCalculator:
    source: "bilby.core.prior.PriorDict"
    cosmology: "Cosmology" = cosmo.Planck15

    def __post_init__(self):
        """Do v0 calculation here"""
        pass

    def reweight_samples(self, recovered_parameters, target):
        """Calculate weights here.

        Only do this if you have a target to reweight by, since
        otherwise the weights will all just be ones and you're
        wasting time on two different `.prob` calculations
        """
        pass

    # since this is all the class is meant to do, probably
    # simpler to make it a callable rather than give this
    # method some redundant name
    def __call__(self, recovered_parameters, n_injections, target = None):
        recovered_parameters = do_list_dict_transpose(recovered_parameters)

        if target is not None:
            weights = self.reweight_samples(recovered_parameters, target)
            mu = np.sum(weights) / n_injections
            variance = np.sum(weights**2) / n_injections**2
        else:
            mu = len(recovered_parameters) / n_injections
            variance = mu**2

        # do the rest of calc

alecgunny:
And now that I think about it, this calculator should probably be initialized during the initialization of the VT plot object and saved as an attribute for future use, since we'll want to reuse it to recalculate the sensitive volume any time we change the mass input values

alecgunny:
The way the data gets plotted is hard to see as well, for example
![image](https://user-images.githubusercontent.com/14932242/218803412-01df673f-3b47-434f-a4bb-892e0b19e600.png)

alecgunny:
[...]

PR #291, opened 2023-02-14
Fixing bugs in vizapp

alecgunny:
I've started to apply some of the changes outlined by #290 , but there's still lots that needs to be done
- [x]  Refactor sensitive volume calculation and debug why initialization takes so long
- [x] Move `redshift` calculation to initialization of `Foreground` object as one of its attributes
- [ ] Make points on VT plot more visible


alecgunny:
Moved the `redshift` computation to the front of the VT calculation loop, but this still takes long enough that we won't want it to be done every time we change our mass distribution, so it should probably be moved to the `Foreground` initialization to pay the cost once fully up front.

@EthanMarx I know we talked about some other fixes/additions we would like the vizapp to have yesterday but I can't seem to think of them now. Any additional to-do's this PR should have?

EthanMarx:
I don't think anything for this PR. I'll open up issues for the other items.

EthanMarx:
Should we get these bug fixes merged in and address the other points separately?

alecgunny:
Yeah I think so, having this be so slow is a real bummer when you have some results you want to analyze

EthanMarx:
I think this just needs to be rebased on main to pull the `isort` update

alecgunny:
Yeah I have a couple more slight optimizations I've made locally, I'll get things fixed and then push to get this merged ASAP

wbenoit26 (approved):

PR #292, opened 2023-02-14
FP16 inference

### LINKED ISSUE #276: Improve inference throughput by returning to FP16 inference for the neural network

alecgunny:
Separating the preprocessor and neural network at export time so that the NN can be exported at reduced precision. Closes #276 . Inferring the inference `kernel_length` from the saved preprocessor weights at export time, so this argument can be removed.

Right now I'm topping out single GPU throughput at around 8,000 inf/s at a batch size of 128 on a V100. I've done some benchmarking with TensorRT's python APIs, and the max I can get it to is 10,000 inf/s, which improves on the full precision throughput of ~5,800 inf/s. So once you factor in the other models on the GPU eating up resources, this sounds about right.

The issue is that with older versions of the code (but with the same size model), I was able to top out single GPU throughput at around 25,000 inf/s. The possible disparities I can think of are
- TensorRT performance actually degrading over time
- BatchNorm -> GroupNorm slowing things down. This might seem unlikely, but Conv -> BatchNorm -> ReLU is a sufficiently common chain of operations that many frameworks (TensorRT included) have fused operators that can execute these really quickly in a single kernel call.

I'm going to spend some time today benchmarking these different possibilities to understand what's going on here and if there's a way we can recover some of this performance, so marking as a draft until this is worked out.

- [x] Sort out cause of throughput drop between older versions of the code and now
- [x] Finish updating scaling tests to account for different numbers of instances of NN and preprocessor

### DIFF
--- a/libs/architectures/bbhnet/architectures/resnet.py
+++ b/libs/architectures/bbhnet/architectures/resnet.py
@@ -55,10 +55,10 @@ class ChannelNorm(torch.nn.Module):
         # roll the mean and variance into the
         # weight and bias so that we have to do
         # fewer computations along the full time axis
-        var = (sq_mean - mean**2 + self.eps) ** 0.5
-        weight = self.weight / var
-        bias = self.bias - weight * mean
-        return bias + x * weight
+        std = (sq_mean - mean**2 + self.eps) ** 0.5
+        scale = self.weight / std
+        shift = self.bias - scale * mean
+        return shift + x * scale
 
 
 def get_norm_layer(groups: Optional[int] = None) -> nn.Module:
--- a/libs/architectures/bbhnet/architectures/resnet.py
+++ b/libs/architectures/bbhnet/architectures/resnet.py
@@ -44,7 +44,9 @@ class ChannelNorm(torch.nn.Module):
         # reshape back to full channel dimension
         if self.num_groups != self.num_channels:
             mean = torch.stack([mean, sq_mean], dim=1)
-            mean = mean.reshape(-1, 2, self.num_group, self.channels_per_group)
+            mean = mean.reshape(
+                -1, 2, self.num_groups, self.channels_per_group
+            )
             mean = mean.mean(-1, keepdims=True)
             mean = mean.expand(-1, -1, -1, self.channels_per_group)
             mean = mean.reshape(-1, 2, self.num_channels, 1)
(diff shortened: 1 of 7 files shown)


alecgunny:
It seems that the main source of this disparity was moving from BatchNorm to GroupNorm, which is poorly optimized in TensorRT. I was able to get around this by implementing a custom version of GroupNorm that does some reshape magic. This implementation costs us about a 15% haircut in training throughput (3000 samples/s to 2500 samples/s), but in exchange we get over 2x improvement in inference throughput (see attached graph). Considering inference is where our bottleneck lies (and will only get worse), I'm prepared to accept this tradeoff.

The alternative is that we train with Torch's built-in GroupNorm, then swap it our for our custom one at export time. This will allow us to get the best of all throughput worlds (though obviously still not as fast as BatchNorm), but at the cost of introducing some noise due to precision. Initial tests show that taking the weights from a trained torch GroupNorm layer and moving them to this custom layer results in ~99.9% of output values agreeing to within a relative tolerance of 1e-6, but it will need to be checked whether this accumulated error across the whole network impacts performance. Because this will be at least slightly intensive, my suggestion will be to merge this implementation (once tests are completed) then leave that for a future PR.

image

libs/architectures/bbhnet/architectures/resnet.py
```diff
@@ -11,14 +11,61 @@
...
+            )
+            mean = mean.mean(-1, keepdims=True)
+            mean = mean.expand(-1, -1, -1, self.channels_per_group)
+            mean = mean.reshape(-1, 2, self.num_channels, 1)
+            mean, sq_mean = mean[:, 0], mean[:, 1]
+
+        # roll the mean and variance into the
+        # weight and bias so that we have to do
+        # fewer computations along the full time axis
+        var = (sq_mean - mean**2 + self.eps) ** 0.5
```
wbenoit26, line 58: Nitpick, but the variance is this without the square root, right?

wbenoit26 (approved):

libs/architectures/bbhnet/architectures/resnet.py
alecgunny: ah yes this is a fair point, forgot to rename the variable when I combined this into one line. I'll update it for the sake of completeness

PR #294, opened 2023-02-17
Vt vizapp improvements

rafia17:
Added the VT table from GWTC-3 catalog document for comparison


EthanMarx:
This is awesome, thank you. Could you add the units somewhere? Also, I want to make sure, have you normalized by the livetime of each analysis (i.e. divide the number in the plot by the livetime) ? These should be quoted in the catalog paper somewhere.

EthanMarx (approved):
Thanks for doing this!

PR #295, opened 2023-02-17
Downweight glitches

### LINKED ISSUE #288: Downweight likelihood of injecting on top of a glitch.
Rather than doing injections during training independent of glitch insertion, give glitches a label so that we can condition the probability of doing injections on whether each sample has a glitch. This way we downweight the likelihood of this happening and make the amount of downweighting a hyperparameter to search over.

alecgunny:
Closes #288 . Running pipeline now to see effect on training
- [x] Write tests for downweighting


projects/sandbox/train/train/data_structures.py
```diff
@@ -76,12 +76,47 @@ def __next__(self) -> Tuple[torch.Tensor, torch.Tensor]:
...
+            glitch_prob = kwargs.pop("glitch_prob")
+            downweight = kwargs.pop("downweight")
+            prob = kwargs.pop("prob")
+        except KeyError:
+            self.downweight = 1
+        else:
+            # not to sound like Fermat but I have a
+            # derivation of this somewhere that I
+            # can't quite find at the moment
+            prob = prob / (1 - glitch_prob * (1 - downweight)) ** 2
```
wbenoit26, line 90: This step is so that we still have the same expected number of waveforms?

alecgunny: Exactly. If you let `p` be the desired expected number of waveforms and `p_0` be the new probability for samples that _don't_ have a glitch inserted, with `d` being the downweight factor and `g` being the glitch insertion probability, you want

$$p = p_0(1 - g)^2 + 2(p_0d)g(1-g) + (p_0d^2)g^2$$

$$p = p_0((1 - g)^2 + 2(1 - g)(dg) + (dg)^2)$$

$$p = p_0((1-g) + gd)^2 = p_0(1 + g(d - 1))^2$$

$$p_0 = \frac{p}{(1 + g(d - 1))^2}$$

wbenoit26: Nice, this all makes sense then

wbenoit26 (approved):

Issue #298, opened 2023-02-17
Parallelize waveform generation


alecgunny:
Once #287 is merged, our waveform generation will take substantially longer due to the fact that the vast majority of our candidate injections will be rejected. It will be worth parallelizing this via condor so that we can iterate on which priors we use with some level of efficiency. A potential scheme could look like

```python
from typing import Sequence, Tuple

import h5py
from typeo import scriptify

def calc_waveforms_per_segment(start, stop, spacing):
    """
    Compute how many waveforms can be injected in a
    given segment if they're spaced some distance apart.
    Maybe account for some jitter and edge buffer as well
    """
    return


@scriptify
def main(
    start: float
    stop: float,
    spacing: float,
    output_fname: str,
    **kwargs  # what else?
):
    """
    Generates the waveforms for a single segment
    """
    n_waveforms = calc_waveforms_per_segment(start, stop, spacing)
    return


def calc_shifts_required(segments, Tb, shift):
    """
    Based off of the lengths of the segments and the
    amount of data that will need to be sloughed off
    the ends due to shifting, calculate how many shifts
    will be required to achieve Tb seconds worth of background
    """
    return


def as_completed_for_condor(job_ids):
    """
    Do subprocess calls to condor_q and parse the outputs
    to decide which of the indicated job_ids have completed
    """
    while job_ids:
        completed, job_ids = parse_condor_q_output(job_ids)
        yield from completed


# new argument to scriptify which tells it to pass any
# unused command line arguments to *args as strings,
# but to enforce that these arguments are parsable by
# the indicated function
@scriptify(args=main)
def deploy(
    segments: Sequence[Tuple[int, int]],
    spacing: float,
    sample_rate: float,
    Tb: float,
    shift: float  # the increment to shift by
):
    shifts_required = calc_shifts_required(segments, Tb, shift)
    max_shift = shift * shifts_required

    shift_map = {}
    for start, stop in segments:
        for i in range(shifts_required):
            arguments = " ".join(args)
            arguments += f"--start {start} --stop {stop - max_shift} --spacing {spacing}"

            # Condor should have some environment variable syntax
            # to update this dynamically for each job
            arguments += f"--output-fname /path/to/tmpfiles/{job_id}.h5"

            # build a sub file that runs `main` as the executable
            # with the indicated arguments and return the submitted job id
            job_id = launch_a_job(arguments)
            shift_map[job_id] = shift

    job_ids = shift_map.keys()
[...]

PR #305, opened 2023-02-19
Parallelize timeslide waveforms

EthanMarx:
Minimum working example of waveform parallelization as described in #298 

- [ ] Unit tests on `deploy`
- [x] Fix subfile getting written with indents
- [x] Concatenation of hdf5 files at the end - use `condor_watch_q` like pyomicron to wait for jobs to finish?
- [ ] Incorporation of new `typeo` arg will clean things up
- [x] Running into errors with `condor` due to how `pinto` modifies `LD_LIBRARY_PATH`


EthanMarx:
Re this condor issue: `/usr/lib64/libcondor_utils_9_0_17.so: undefined symbol: _ZN7classad7ClassAdC1Ev`

It looks like pre-pending the conda envs `lib` is confusing the linker. Prepending`/usr/lib64` seems to be a bandaid solution to the problem, but I'm not sure if this will in turn mess with the torch/cuda shared libraries this solution was initially meant to address.

alecgunny:
Take a look at ML4GW/typeo#15 when you have a chance and let me know if the `args` option looks like it might help whatever it is you're building. Haven't thought as much about the nuts and bolts but figure something like this might help with what we're looking for

EthanMarx:
This would definitely help clean up the condor file argument construction

projects/sandbox/datagen/datagen/utils/timeslide_waveforms.py
```diff
@@ -0,0 +1,131 @@
...
+        segments: A list of tuples of the start and stop times of the segments
+        Tb: The amount of background data to generate
+        shift: The increment to shift the data by
+
+    Returns the number of shifts required to achieve Tb seconds of background
+    """
+
+    livetime = sum([stop - start for start, stop in segments])
+    n_segments = len(segments)
+    shifts_required = 0
```
wbenoit26, line 32: Wouldn't save you much, but you could start at `(livetime - 1) // Tb + 1`

Or, you could solve analytically, but that's probably more complicated and less readable.

wbenoit26: Rather, `(Tb - 1) // livetime +1`

EthanMarx: Yeah calculating it analytically required solving a quadratic, I got lazy and just did it this way. It's more robust and probably faster analytically so can do that.

EthanMarx: On second thought, I think this is simpler, shouldn't eat up that much time, and is only done once.

wbenoit26: Yeah, fully agree, the loop is the way to go

alecgunny: One thing I was thinking about today as I was putting the inference stuff together is that as far as I can tell, the main motivation of having the timeslides be the same length was to keep filenames consistent and make it easier to make a correspondence between the files in different shift directories. Since we don't have to do that anymore, do we need to enforce this constraint, or can the 0-shift sequence just be longer than the other sequences?

I guess that will make some headache on dealing with the saved sequences of the _inferred_ data, but if we make the cluster distributions in the inference script as well that largely mitigates this. In fact, there's probably an argument that you don't actually need to save the network outputs at all: given the model weights and the input data you can always reconstruct them (i.e. we could do inference live in the vizapp).

EthanMarx: I agree, my only concern is we would be breaking many projects (e.g. analysis, vizapp) with this one change , but if they are going to change anyways then might as well rip the band-aid off?

alecgunny: I think I'm in favor of ripping it but we can discuss more during our sync today. Would it maybe make sense to make this PR against the `inference-overhaul` branch and have it plug in with those changes, that way we don't have a broken version of `main`?

EthanMarx: Definitely I realized yesterday I PR'd against the wrong branch

EthanMarx:
Think this is worth reviewing and merging at this point. Integration with the rest of the inference scripts can be done via other PR's before the inference-refactoring branch is merged into main.

alecgunny (approved):
Agreed that we should just merge this and then we can sort out the pieces that need to get updated to integrate with inference later

wbenoit26 (approved):

EthanMarx:
@alecgunny Tests are now failing here due to the latest `pinto` fix (i.e. looking for libs in the base conda env). Is there syntax for turning on the ld_lib_path appending from the command line?

projects/sandbox/datagen/datagen/utils/timeslide_waveforms.py
```diff
@@ -0,0 +1,135 @@
...
+        total_livetime = (livetime - n_segments * max_shift) * shifts_required
+        if total_livetime < Tb:
+            shifts_required += 1
+            continue
+        break
+
+    return shifts_required
+
+
+def merge_output(datadir: Path):
```
alecgunny, line 43: Ran some quick tests, first big thing that jumps out at me is that if you can have some way of knowing how many _total_ injections you'll be writing and allocate the memory in the datasets up front, that's about twice as fast as iteratively re-allocating memory in the dataset objects

alecgunny (commented):
See note about hdf5 write

Issue #277, opened 2023-02-07
Write inference script as a single run of a particular shift across all segments to parallelize along shift dimension


alecgunny:

EthanMarx:
I'm looking at the current `infer` project and I don't think much will have to change.
We can simply point the `SequenceManager` directly at a `TimeSlide` directory instead of the `datadir` that contains all of the `TimeSlide` dirs

EthanMarx:
Forgot that we also wan't to be doing injections on the fly, and likely won't be saving any data to disk.
Here's my first go at pseudo code of what this might look like

```python
def infer(
    waveform_path: Path, # path to omnibus waveforms file
    shift: float,
    segments: List[Tuple[float, float]],
    triton_server_ip: str, # already spun up on SDSC or something
    ...
):

    waveforms = load_waveforms(waveform_path)

    client = InferenceClient(triton_server_ip, model_name, model_version)

    # something analogous to the SequenceManager we have now
    sequence_manager = SequenceManager(write_dir, client, ...)

    for segment in segments:
        # load data and timeshift
        background = load_data(segment, channel, ifos)
        background = np.roll(background, shift * sample_rate)

        # add to inference queue
        sequence_manager.add_to_queue(background, field="background")
       
        # slice out waveforms, inject, and add to inference queue
        waveforms = find_waveforms_in_segment(segment, waveforms)
        injected = inject_waveforms(waveforms, background, times)
        sequence_manager.add_to_queue(injected, field="injected")

    events = defaultdict(ClusterDistribution)
    while not sequence_manager.is_done():
        # instead of writing nn outputs to disk, the callback will be clustering events
        future = client.get()
        if future is None:
            time.sleep(1e-2)
            continue

    
        while not future.done():
            sequence_events, field = future.result()
            events[field].append(sequence_events)
```

alecgunny:
I think something like this will make sense, but the point of the `SequenceManager` was really just to make parallelization easier. Since we're parallelizing via Condor now, I think we can drop this and just manage each sequence explicitly in serial. I'll see if I can mock something based off of this up this afternoon.

EthanMarx:
I see. If it doesn't add too much complexity, is it not worth trying to parallelize within each condor job as well?

alecgunny:
[...]

PR #308, opened 2023-02-22
overhauling inference picture to single script

alecgunny:
Addresses #277 by making inference a single serial run across all files for a single shift. Includes a function `infer` which would fit into a more robust inference application down the line by parallelizing across all files and shifts.

Using `TimeSeriesDict`s for hdf5 I/O for the time being, which breaks with some of our existing infra but I think should simplify things, especially since the integration is wrapped into here anyway. I have also dropped normalization since recent results suggests it doesn't help much but this should be easy to add back in if we'd like to.

(Note that we can still keep an `analysis` project for re-integrating/normalizing/clustering already computed predictions using novel methods for doing any of those steps, but just not include it in the `pinto` pipeline).

Remaining questions:
- [x] Since we're doing integration here, do we just want to fit a `ClusterDistribution` in the callback and then aggregate them all in the main thread? I think this will probably make the most sense.
- [x] How do we fit in injections? Are they done alongside their corresponding background files (would be necessary if we were doing normalization), or do we assume that they will be handled by a separate job?
- [x] No `mldatafind` integration right now, so could be memory-limited for long segments. Would make the logic for shifting and concatenating batches across chunks more complex, but would probably be worth it longer term.

TODOs
- [x] ~~Test on a single shift value with server running~~ ran on 0 lag background using latest production model, all raw and integrated outputs matched those from existing inference/analysis scripts exactly.
- [ ] Rewrite tests
- [x] Download testing segments on head node during timeslide waveform generation. Use this to replace existing `generate-timeslides` script
- [x] Put sections/arguments for all the new scripts in the `sandbox` pyproject.toml


EthanMarx:
1. Fitting the ClusterDistributions makes sense since we already have all the info they need
2. Performing and analyzing the injections would fit nicely in `infer`, but would kill its generality for a larger inference application. The alternative would be to perform the injections in `main`, and then the `fname` argument of `infer` could be generalized to `Optional[path, np.ndarray]`. We could then either read the data from disk or pass it directly
3. We can keep what we currently do in the `timeslide` project, which is chunk segments if they are longer than a specified duration.

alecgunny:
Most of the required infra is in there at this point. All that remains is:
- [ ] Tests for all the new modules to verify functionality
- [ ] `events` API in `analysis` library that gets imported here but doesn't exist yet
- [ ] Integration with Ethan's condor code

alecgunny:
This has become an enormous undertaking, but the pipeline is now working. The structure of the data directory is now:
```
train/
    background/
        background-<t0>-<duration>.h5
    glitches.h5
    waveforms.h5
test/
    background/
        background-<t0_0>-<duration_0>.h5
        background-<t0_1>-<duration_1>.h5
        ....
    waveforms.h5
```
Couple things to note:
- Waveform generation is bottlenecked by dataloading within each client, and on aggregation of each client's results. This is not the end of the world since you only need to generate a new dataset once in a while (and hopefully we've nailed it such that it's a long while), but worth thinking about optimizing this down the line. Possibly moving all the data generation into a single step so they all run in parallel and the waveform generation can take advantage of a pre-comptued train background PSD (which will be way smaller than the segment needed to compute that PSD on each waveform generation job)
- Viz app won't work with this new set up, but hopefully the new `bbhnet.analysis.ledger.events` API, and the removal of multiple normalizations, should only simplify the vizapp code. Given that we want to make a lot of adjustments to it anyway, maybe this is all better handled in a follow up PR.

I'm going to mark this as ready for review so that folks can begin taking a look and we can see which tests fail. I think the final list of to-dos is:
- [ ] The `num_injections` is getting wildly inflated during event recovery: the number associated with the test `waveforms.h5` is correct, but the one that ends up in `foreground.h5` at the end of inference is a few orders of magnitude too high.
- [ ] There will definitely be tests that I have forgotten to strip out for no-longer-existing modules, calls to tests in no-longer-existing libraries in the CI workflow that need to be removed, and whole projects like `analysis` that still need to be stripped out (and whose tests will likely fail).
- [ ] I have given everything in here reasonably thorough eyeball tests: looking at the distributions of the masses of the test waveform injections, running through inference of a couple test segments and making sure that the data comes out right ( concatenated background matches the standalone file, `injection - background` comes out to clean waveforms) and that the recovered detection statistics and times and correct, but more extensive unit testing to verify everything will be important. I don't know that we need to be exhaustive before this gets in or before we start using it, but more than what's in there now is probably necessary.

For reviewers, this is enough of a behemoth that it might just be easier to go to the branch itself and peruse the code, since the diffs will be enormous and confusing. Would have been smart to handle this in smaller chunks, but things escalated as they tend to do.

EthanMarx:
I opened up a PR against your repo to address some issues. I noticed that a few of the `ledger` test were failing. Not sure If theres a commit or two missing here.

PR #309, opened 2023-02-23
Added chirp mass constraint to prior constraint function

wbenoit26:


wbenoit26:
@EthanMarx is this good for what you want, or should we also bump up the minimums on the masses?

EthanMarx:
This implements the ability to apply a chirp mass constraint, but not an actual chirp mass constraint, correct?

wbenoit26:
Check that you're looking at the latest commit - I forgot to add the actual `Constraint` the first time

EthanMarx:
Could you plot the resulting m1 / m2 distributions and also determine the maximum signal duration? 
We could also just only bump up the minimum masses to 10 and in that case we'll know the max is ~ 1.7

wbenoit26:
image

Here's the mass distribution. The signal duration is just determined by chirp mass, right? A 10/10 merger has a chirp mass of ~8.7, so constraining chirp mass to > 10 should have a max signal duration of < 1.7 s

EthanMarx:
Yeah exactly. I was being lazy and didn't calculate, wanted to make sure the math was done haha.

wbenoit26:
Ha, excellent. For future reference, the explicit max signal duration with this constraint. should be about 1.3-1.4 seconds

EthanMarx (approved):
Was that `bilby` constraint fixed by the `scipy` constraint?

wbenoit26:
Yeah, I don't have the full picture of what the issue was, but `astropy` was entering through `bilby`, so we constrained both to avoid conflicts.

PR #310, opened 2023-02-24
Split glitches properly

EthanMarx:
1. train/val split of glitches is done in time to ensure no contamination of training glitches in the validation set
2. Generalize glitch split to multi ifos


projects/sandbox/train/train/train.py
```diff
@@ -47,6 +47,7 @@ def main(
     kernel_length: float,
     sample_rate: float,
     batch_size: int,
+    train_stop: float,
```
alecgunny, line 50: Oh wow I didn't realize that we stopped passing the train start time to `train.py`, interesting... In either case, would it maybe make more sense to use that, then infer this like `train_stop = train_start + (1 - valid_frac) * background.shape[-1] / sample_rate`?

EthanMarx: Since we started querying glitches before the training background start time to increase numbers, I figured just splitting from the end of the training time made the most sense.

EthanMarx: I'm now realizing `train_stop` includes validation. Will have to update that logic

EthanMarx: Renaming variables for clarity and calculating this as below:
```
valid_start = (1 - valid_frac) * (train_val_stop - train_val_start) + train_val_start
train_gitches = glitches[times <= valid_start]
valid_glitches = glitches[times > valid_start]
```

projects/sandbox/train/train/train.py
```diff
@@ -47,10 +48,9 @@ def main(
     kernel_length: float,
     sample_rate: float,
     batch_size: int,
-    mean_snr: float = 8,
-    std_snr: float = 4,
-    min_snr: Optional[float] = None,
-    highpass: Optional[float] = None,
+    highpass: float,
+    train_val_start: float,
```
alecgunny, line 52: I think the issue with making the validation start and stop times the parameters here is that it would need to change these with each choice of `valid_frac` (and would really obviate that parameter altogether). I think if you just specify the start of the training period, you can infer the duration from the length of the background and the sample rate, and then from the `valid_frac` you can infer where the valid period begins.

In fact, now that I think about it, we save `t0` with the training data, so we can probably infer all of this without any additional parameters, right?

alecgunny (commented):
See note about additional parameters. I think we can do this all implicitly, or am I over-simplifying things?

projects/sandbox/train/train/train.py
EthanMarx: `train_val_start` and `train_val_stop` refers to the start time of the combined training and validation period. The logic for calculating the validation start is still there. I renamed things for clarity since `train_start` and `train_stop` actually referred to training _and_ validation start and stop

EthanMarx: Okay took care of this, give it another look. 

Background is now saved in one file with the `ifo` as the dataset key, and `t0` as an attribute. `ifos`, `train_val_start` and `train_val_stop` are inferred from this file.

projects/sandbox/datagen/datagen/scripts/background.py
```diff
@@ -110,8 +110,10 @@ def main(
                 f"The background for ifo {ifo} contains NaN values"
             )
 
-        with h5py.File(datadir / f"{ifo}_background.h5", "w") as f:
-            f.create_dataset("hoft", data=data)
-            f.create_dataset("t0", data=float(segment[0]))
+        with h5py.File(datadir / "background.h5", "w") as f:
```
alecgunny, line 113: God bless you for making this change

alecgunny (approved):
Lot of great updates in here, thanks for putting this together

PR #313, opened 2023-03-02
Reverted back to ConditionalPriorDict for `end_o3_ratesandpops`

wbenoit26:
This also switches back to `UniformComovingVolume`.

Histogram of `mass_1` distribution with desired distribution overlaid:
image


EthanMarx:
I think the biggest change is the following:
```python
prior["theta_jn"] = Sine()
```

The other changes I made are outlined in this prior: 
https://git.ligo.org/lscsoft/bilby/-/blob/master/examples/gw_examples/data_examples/GW150914.prior

Basically, making `psi` and some of the spin parameter uniform. Let me know if there are other questions

wbenoit26:
Double-checking before I merge - @EthanMarx does this match your changes?

EthanMarx:
Yup that looks good!

PR #320, opened 2023-03-28
Validation auroc fix

EthanMarx:
Rescales validation waveforms with SNRs below some threshold to the threshold


projects/sandbox/pyproject.toml
```diff
@@ -100,20 +100,20 @@ outdir = "${base.basedir}/training"
...
 # optimization parameters
 batch_size = 512
 batches_per_epoch = 200
 max_epochs = 200
-max_lr = 4e-3
-lr_ramp_epochs = 75 
-weight_decay = 1e-5
+max_lr = 0.21355 
+lr_ramp_epochs = 83 
+weight_decay = 6.28608e-06
```
wbenoit26, line 116: These params should be set back to their defaults.

wbenoit26 (changes requested):
Looks good other than the changes to the config. How many signals did we end up mapping to an SNR of 6?

projects/sandbox/pyproject.toml
EthanMarx: Might be a good time to update to the best params from the hp search. Can you share those here?

wbenoit26: Good idea:
```
waveform_prob: 0.470644, 
glitch_prob: 0.420902, 
max_lr: 0.000181352, 
lr_ramp_epochs: 92, 
weight_decay: 2.81536e-06,
glitch_downweight: 0.650283
```

wbenoit26 (approved):

Issue #317, opened 2023-03-11
Training augmentation: swap one channel of waveforms


alecgunny:
Before injection, randomly select some subset of the projected waveforms to swap one of their interferometer channels and _not_ get marked as signal. This could help regularize the network to learn that it's not sufficient that there just be a BBH-looking event in each IFO: the events have to have matching templates as well.

EthanMarx:
So swap a single IFO's channel from two kernels with injections so that hanford and livingston have BBH signals with different parameters in them?

alecgunny:
Yeah exactly. I think the only change to how the injection transform gets built is that you would need something like a `p_aug` argument to the class (just what it needs, another parameter) and just divide by this when calculating the new injection probability due to downweighting (since `p_aug` fraction of your injections will end up not being marked as signal, you need to increase the probability of originally selecting samples for injection accordingly)

alecgunny:
Closed by #328

PR #321, opened 2023-03-28
Channel swap augmentor

EthanMarx:
Augmentor that swaps channels of waveforms and labels them as noise.
Addresses #317


projects/sandbox/train/train/data_structures.py
```diff
@@ -202,3 +213,29 @@ def forward(self, X, y):
...
+        self.frac = frac
+
+    def forward(self, X):
+        num = int(X.shape[0] * self.frac)
+        num = num if not num % 2 else num + 1
+        indices = torch.randperm(X.shape[0])[:num]
+        channel = torch.randint(X.shape[1], size=(num // 2,))
+        print(X.shape, channel.shape, num)
+        X[indices[-num // 2 :], channel] = X[indices[: num // 2], channel]
+        X[indices[: num // 2], channel] = X[indices[-num // 2 :], channel]
```
wbenoit26, line 239: This won't change anything, because of the previous line. We'll need to make a copy of `X` first.

EthanMarx: Good call

projects/sandbox/train/tests/test_data_structures.py
```diff
@@ -338,3 +339,31 @@ def test_signal_reverser(flip_prob, rvs, true_idx):
...
+    ), patch("torch.randint", return_value=channels):
+        X, indices = tform(X)
+
+    X = X.cpu().numpy()
+    indices = indices.cpu().numpy()
+
+    # this seems like a circuitious way to test this
+    assert all(indices == np.arange(num))
+    assert (
+        X[indices[: num // 2], channels] == X[indices[-num // 2 :], channels]
```
wbenoit26, line 368: This also needs to be compared to a copy, though I agree, it does feel circular as-is. I don't mind it too much, as it's just the most direct check of a swap that you could possibly have.

Also, there's a print statement a few lines up.

projects/sandbox/train/train/data_structures.py
EthanMarx: Arrived on an implementation that removes the need to copy

wbenoit26: Nice, I like it. Solves the circularity of the testing, too.

projects/sandbox/train/train/data_structures.py
```diff
@@ -118,8 +125,12 @@ def forward(self, X: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
             max_center_offset=self.trigger_offset,
             coincident=True,
         )
+
+        waveforms, indices = self.channel_swapper(waveforms)
         X[mask] += waveforms
 
+        # make targets negative if they have had a channel swapped
+        mask = mask[np.where(mask)[0]][indices] = 0
```
wbenoit26, line 133: Can you explain this line? I'm having trouble parsing it.

EthanMarx: The first `mask = ` is a typeo. Would that clarify any confusion?

wbenoit26: Ah, that helps. The shapes of everything is breaking my brain - why the `[0]`?

wbenoit26: Got it, okay, cool

projects/sandbox/train/train/utils.py
```diff
@@ -123,11 +124,13 @@ def prepare_augmentation(
                 plus=valid_plus,
                 cross=valid_cross,
             )
+            print(valid_injector)
+            print("here")
         else:
             valid_injector = None
+            print(valid_injector)
```
wbenoit26, line 131: More print statements here

projects/sandbox/train/train/data_structures.py
EthanMarx: But in short, `mask` is a boolean array which refers to samples of the background where we will inject waveforms. Due to the channel swapping, some subset waveforms have a channel swapped, and therefore need to be labeled as noise. `Indices` indexes the subset of mask where mask is `True`. These are waveforms that have been swapped, and thus their label should be set to 0.

wbenoit26 (approved):

EthanMarx:
@wbenoit26 I should probably write some tests of the full `BBHnetWaveformInjection` object with `swap_frac` before merging.

wbenoit26:
Oh okay, my bad, I thought this was ready to go

projects/sandbox/train/train/data_structures.py
```diff
@@ -8,6 +8,33 @@
...
+        super().__init__()
+        self.frac = frac
+
+    def forward(self, X):
+        num = int(X.shape[0] * self.frac)
+        indices = None
+        if num > 0:
+            num = num if not num % 2 else num + 1
+            channel = torch.randint(X.shape[1], size=(num // 2,)).repeat(2)
+            indices = torch.randperm(X.shape[0])[:num]
```
alecgunny, line 31: Not a big deal, especially if this already works, but I actually don't think we need to be super diligent about randomness here due to the fact that gradient descent just averages over the batch:
- You can swap the same channel each time, it's functionally the same as swapping the other channel then swapping the batch elements
- You can just swap the first `num / 2` indices with the next `num / 2` indices since our waveform sampling and background sampling are both uniform (i.e. have no dependence on their position within a batch)

But this is obviously more general, so maybe it's worth keeping.

projects/sandbox/train/train/data_structures.py
```diff
@@ -118,8 +153,13 @@ def forward(self, X: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
             max_center_offset=self.trigger_offset,
             coincident=True,
         )
+
+        waveforms, indices = self.channel_swapper(waveforms)
         X[mask] += waveforms
 
+        # make targets negative if they have had a channel swapped
+        if indices is not None:
+            mask[np.where(mask)[0]][indices] = 0
```
alecgunny, line 162: You should just be able to go `mask[mask][indices] = 0`. In either case, I think you maybe mean to use `torch.where`

alecgunny (commented):
See couple minor notes. Only the `np.where` I think might break things

projects/sandbox/train/train/data_structures.py
EthanMarx: This makes things much clearer so all for it

projects/sandbox/train/train/data_structures.py
EthanMarx: Ah masking a mask, good call

projects/sandbox/train/train/data_structures.py
EthanMarx: Can I ask why not the same argument for the `Signalnverter` and `SignalReverser`?

PR #322, opened 2023-03-29
Reverted to LogNormal(15, 15, 1) distribution

wbenoit26:


wbenoit26:
I'm confused as to why the pre-commit check is taking so long here

PR #326, opened 2023-04-04
Sv fix

EthanMarx:
1. Sensitive volume calculator now expects source frame quantities
2. Adjusts vizapp accordingly

### DIFF
--- a/libs/analysis/bbhnet/analysis/sensitivity.py
+++ b/libs/analysis/bbhnet/analysis/sensitivity.py
@@ -146,11 +146,11 @@ class SensitiveVolumeCalculator:
             mu = len(recovered_parameters) / num_injections
             variance = len(recovered_parameters) / num_injections**2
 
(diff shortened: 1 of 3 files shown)


libs/analysis/bbhnet/analysis/sensitivity.py
```diff
@@ -154,10 +146,11 @@ def __call__(
             mu = len(recovered_parameters) / num_injections
             variance = len(recovered_parameters) / num_injections**2
 
-        variance -= mu**2 / num_injections
-        v = mu * self.volume
-        variance *= v**2
+            variance -= mu**2 / num_injections
```
wbenoit26, line 149: The lines from here onward are indented an extra time

EthanMarx: good catch

wbenoit26 (approved):

PR #328, opened 2023-04-11
`ChannelMuter` and `ChannelSwapper`

EthanMarx:
Adds objects that will swap waveform channels and mute waveform channels and label those as noise.


EthanMarx:
reopened version of #321

projects/sandbox/train/train/data_structures.py
```diff
@@ -76,27 +131,37 @@ def __next__(self) -> Tuple[torch.Tensor, torch.Tensor]:
...
+    ):
+        self.downweight = downweight
+        self.prob = prob
+        # not to sound like Fermat but I have a
+        # derivation of this somewhere that I
+        # can't quite find at the moment
+        prob = prob / (1 - glitch_prob * (1 - downweight)) ** 2
+        # account for the fact that some waveforms will have a channel
+        # swapped with another waveform and labeled as noise
+        prob = prob / (1 - (swap_frac + mute_frac))
```
wbenoit26, line 152: I think this should be `prob = prob / (1 - (swap_frac + mute_frac - swap_frac * mute_frac))` because both can happen

EthanMarx: agreed

wbenoit26 (approved):

PR #330, opened 2023-04-18
Integrating ledger API into vizapp

alecgunny:
- [ ] Create submodule of `priors` with all the targets of interest for analysis
- [ ] Pre-compute probability of parameters wrt to source and all targets during data generation
- [ ] Load weights along with parameters via `Ledger` objects when starting up vizapp
- [ ] Implement sensitive volume calculation using loaded weights and `Ledger` objects


alecgunny:
@EthanMarx at this point all the tests are working except some of the training tests that I think have either been fixed in `main` or will need to be updated anyway with your coming fixes, so I think this should be good to go. Most obvious missing tests are (not including the deploy scripts, which will take a lot of clever but annoying mocking)

- `datagen.scripts.background` (deleted because they were way out of date, but obviously very relevant to the redundant background issue we were seeing)
- `infer.main`
- Everything in `vizapp` (which is still incomplete)

EthanMarx (approved):
Looks good. Only comment is that we should open up an issue to document the ledger API thoroughly. Will take a look at failing background tests.

PR #338, opened 2023-05-08
Online training 

EthanMarx:
Begin developing online training code to get the ball rolling. Mostly modeled off the deepclean online training deployment.

### DIFF
--- a/projects/microservice/trainer/trainer/trainer.py
+++ b/projects/microservice/trainer/trainer/trainer.py
@@ -1,0 +1,99 @@
+import time
+from pathlib import Path
+from typing import List
+
+from lal import gpstime
+from mldatafind.io import fetch_timeseries
+from mldatafind.segments import query_segments
+from type.utils import make_dummy
+from typeo import scriptify
+
+from bbhnet.architectures import architectures
+from bbhnet.trainer import train
+
+
+def find_segments(
+    channels: List[str],
+    segment_names: List[str],
+    start: float,
+    cadence: float,
+    duration: float,
+    wait: int = 120,  # length of time dqsegdb takes to update
+):
+    """
+    Iterator that returns training segments
+    """
+    segments = []
+    while True:
+        stop = int(gpstime.gps_time_now())
+        segments = query_segments(
+            segment_names,
+            start,
+            stop,
+            duration,
+        )
+        if segments:
+            # if multiple segments, use most recent segment;
+            # TODO: should we enforce segment duration is
+            # equal to requested train_duration?
+            start, stop = segments[-1]
+
+            data = fetch_timeseries(channels, start, stop, array_like=True)
+            yield start, stop, data
+
+            start += cadence
+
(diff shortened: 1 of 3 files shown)


projects/microservice/trainer/trainer/trainer.py
```diff
@@ -0,0 +1,99 @@
...
+        start,
+        cadence,
+        train_duration,
+    )
+
+    for start, stop, background_data in segment_it:
+        # logic for organizing glitch dataset for training
+        # based on start, stop of training background data
+
+        glitches = get_glitches(start, stop)
```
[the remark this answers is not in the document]
EthanMarx: Agreed. Maybe those parameters and logic should all be handled by the project / service running `pyomicron`. Here we could just read in the current `glitches.h5 `

projects/microservice/trainer/trainer/trainer.py
```diff
@@ -0,0 +1,100 @@
...
+# TODO: add new typeo as submodule
+@scriptify(
+    kwargs=make_dummy(train, exclude=exclude),
+    architecture=architectures,
+)
+def main(
+    segment_names: List[str],
+    channels: List[str],
+    train_duration: float,
+    cadence: float,
```
wbenoit26, line 74: I'm thinking it might make sense to have a cadence for a re-training pipeline (train a new model, build a new test set, run inference over the test set, build a background distribution of network outputs) rather than a cadence for an individual component

alecgunny: FWIW the way I'm handling this in DeepClean is that the training project sets the cadence, then triggers the downstream projects:
- Makes a request to an export service pointing it to the new version of the weights. Export service pushes a new entry to the Triton model repository
- Triton automatically picks up the new entry and serves it up as they "canary" version
- Nothing changes in the inference pipeline: it continues making requests to the canary and production versions of the model on the inference service, but doesn't know or care what they are. All it cares about is that the frames it writes don't have an ASDR>1 (if they do, it'll just write the raw frames instead)
- Monitoring service notices that there's a new canary version of the model, and starts monitoring the canary frames produced by the inference service to ensure that they fall into spec (just ASDR check for now). Once they do for long enough, it sends a request to the export service to move the canary model into production

Obviously this is all overkill for what we need now, but something like it (with an extra evaluation step for building a threshold) would probably work in practice.

Conceptually, the cadence rightfully belongs in the training projects purview, and there will be things that it will have to do to potentially bend that cadence, e.g. if the `ANALYSIS_READY` flag goes off, it will make the decision to stop training until it goes back down. The inference pipeline doesn't need to be concerned with these details, all it needs to know is that the production model that it's making requests to has been appropriately validated. The one difference here is that we don't have something akin to an ASDR to check to ensure that even if a model goes stale, we don't rely on its outdated predictions, so maybe there's some time-based decay in the confidence of our triggers if the production version remains un-updated for a significant period of time.

PR #342, opened 2023-05-15
Batch augmentor

EthanMarx:
Implements new batch augmentor that allows incorporation of SNR curriculum learning (currently just a dummy step that does nothing).

### DIFF
--- a/projects/sandbox/train/tests/Untitled-1.ipynb
+++ b/projects/sandbox/train/tests/Untitled-1.ipynb
@@ -1,0 +1,393 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/home/ethan.marx/miniconda3/envs/bbhnet-datagen/lib/python3.9/site-packages/tqdm/auto.py:21: TqdmWarning: IProgress not found. Please update jupyter and ipywidgets. See https://ipywidgets.readthedocs.io/en/stable/user_install.html\n",
+      "  from .autonotebook import tqdm as notebook_tqdm\n"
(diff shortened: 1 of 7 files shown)


wbenoit26:
This looks pretty nice to me, other than the extraneous notebook. I lost track of the status of your `SnrRescaler` PR in `ml4gw` - is that more or less good to go, and would you prefer to get that merged in and use that here?

EthanMarx:
There will never not be an extraneous notebook from me. 

The `SnrRescaler` PR is coupled with other changes that I'm now not really a huge fan of. I do think the `SnrRescaler` itself is a sufficiently general / useful tool, so maybe we merge that in by itself in a separate PR. 

Still need some unit tests here too

wbenoit26:
Ah, I remember now, yeah. I think integrating the `SnrRescaler` in a separate PR makes sense, as long as there aren't any issues building it into `RandomWaveformInjection`.

Yeah, for sure.

EthanMarx:
I think the idea was the break up / remove `RandomWaveformInjection`. After this PR we won't actually use it in `aframe` (almost said BBHNet there ). The `SnrRescaler` factors out a large component of it (the `.fit` method). I'm not completely sure what else is worth generalizing. Maybe the `sample_responses` function that I have in the `BatchAugmentor`?

wbenoit26:
Isn't that what `waveforms.py` takes care of in your `ml4gw` PR? Are those the other changes that you're not a fan of?

EthanMarx:
Yeah exactly. I'm thinking it might be best to combine the `Sampler` and `Projector`.

wbenoit26:
That seems good to me, maybe with a flag to determine whether calling it returns the raw waveforms or the projected ones

EthanMarx:
I think I would rather keep them separate actually than introduce a flag. Ahh I keep going back and forth.

PR #345, opened 2023-05-17
Changed BBHNet to aframe

wbenoit26:
Wasn't sure how we wanted to capitalize things. I've used `Aframe` in class names, and `aframe` everywhere else. Might want to also use `Aframe` for the beginning of sentences and the VizApp title, but I could go either way.


wbenoit26:
@alecgunny @EthanMarx If there are nothing I missed, could we get this merged in? In case we start getting eyes on our code, I want things to be consistent and not confusing

alecgunny:
@wbenoit26 @EthanMarx depends if the higher priority is getting the batch augmentor/upcoming validation PRs into main, or if it's fine to have those living on different branches that we can run experiments like the HP search off of for now. This will obviously complicate those PRs, but if all we care about is being able to run them, then not having those on `main` is fine. Are the new eyes your concerned about stemming from our application to GraceDB?

wbenoit26:
Exactly, yeah. I've gotten a couple of questions about what exactly we're doing, and it would be good to be able to point people to our repository in addition to just describing things. That's going to happen more often as we start to interact with the wider ecosystem, and I want to make a good first impression (i.e., we have our shit together). I'm fine having things live off of `main`, though I'm not sure how much this change will really complicate things

wbenoit26:
Also, I think we'll keep having reasons to put off this name change, might as well bite the bullet

EthanMarx:
My vote is to get the batch augmentation, snr rescheduling, and validation PRs into main first before the big renaming PR.

At this point we have too many branches that is making it almost impossible to know what is where. So I think we should prioritize consolidating these above features into main first ahead of cosmetic changes

In any case given some of the new libraries / API changes, won't we have to essentially do this all again when we merge in `inference_refactoring`?

wbenoit26:
We've already merged the inference changes - that was why we initially delayed the renaming. My point is that we always have some open PRs that would be messed up by this change, but it needs to happen, so it might as well be now.

wbenoit26:
What's the specific concern with merging now? Rebasing is pretty easy - is it rebuilding the environments that's the problem?

EthanMarx:
Batch augmentor tests are ready so that and the SNR scheduler can be merged in. I say we get those in first

wbenoit26:
I feel like I'm being obstinate, but I'm just confused - why is there so much resistance to merging this in? I don't think order matters: it'll either be 10 minutes to redo the name change after the batch augmenter gets in, or it'll be 10 minutes to rebase the batch augmenter on top of the name change and rename `BBHNetBatchAugmentor` to `AframeBatchAugmentor`.

Like, sure, at this point with the batch augmenter ready to go, it doesn't matter, but we could have had this in a week ago.

EthanMarx:
I was expecting additional unforeseen conflicts that would slow down getting the augmentor / scheduler in that I didn't wan't to deal with right now, but hopefully I'm wrong. 

Let's figure this out on the call later today

alecgunny (approved):

PR #349, opened 2023-05-26
Timeslide validation

alecgunny:
Has been rebased on top of the current `batch-augmentor` branch, will rebase to `main` and add some tests once that gets merged so the diff is more obvious

### DIFF
--- a/projects/sandbox/train/train/train.py
+++ b/projects/sandbox/train/train/train.py
@@ -1,34 +1,19 @@
+import logging
 from pathlib import Path
-from typing import Literal, Optional
+from typing import Optional
 
-import h5py
-import numpy as np
-from train.data_structures import AframeInMemoryDataset
-from train.utils import prepare_augmentation, split
-from train.validation import BackgroundAUROC, GlitchRecall, Recorder, Validator
+from train import utils as train_utils
+from train import validation as valid_utils
+from train.data_structures import (
+    AframeInMemoryDataset,
+    GlitchSampler,
+    SnrRescaler,
+    SnrSampler,
+)
 
 from aframe.architectures import Preprocessor
 from aframe.logging import configure_logging
 from aframe.trainer import trainify
-
-
-def load_background(background_dir: Path):
-    # load in the data from the first segment in the background directory
-    try:
-        background_dataset = next(background_dir.iterdir())
-    except StopIteration:
-        raise ValueError(
-            f"No files in background data directory {background_dir}"
-        )
-    background = []
-
-    with h5py.File(background_dataset, "r") as f:
-        ifos = list(f.keys())
-        for ifo in ifos:
-            hoft = f[ifo][:]
-            background.append(hoft)
-        t0 = f[ifo].attrs["x0"]
-    return np.stack(background), ifos, t0
 
 
 # note that this function decorator acts both to
@@ -46,31 +31,31 @@ def main(
     waveform_dataset: Path,
     outdir: Path,
     logdir: Path,
-    # data generation args
-    glitch_prob: float,
-    waveform_prob: float,
-    glitch_downweight: float,
-    snr_thresh: float,
-    kernel_length: float,
-    sample_rate: float,
+    # optimization args
     batch_size: int,
+    batches_per_epoch: int,
+    snr_thresh: float,
     max_min_snr: float,
-    min_min_snr: float,
     max_snr: float,
     snr_alpha: float,
     snr_decay_steps: int,
+    # data args
+    sample_rate: float,
+    kernel_length: float,
+    fduration: float,
     highpass: float,
+    # augmentation args
+    glitch_prob: float,
+    waveform_prob: float,
+    glitch_downweight: float,
     swap_frac: float = 0.0,
     mute_frac: float = 0.0,
-    batches_per_epoch: Optional[int] = None,
-    # preproc args
-    fduration: Optional[float] = None,
     trigger_distance: float = 0,
     # validation args
     valid_frac: Optional[float] = None,
     valid_stride: Optional[float] = None,
-    monitor_metric: Literal["background", "glitch"] = "glitch",
-    threshold: float = 1.0,
+    max_fpr: float = 1e-3,
+    valid_livetime: float = (3600 * 12),
     early_stop: Optional[int] = None,
     checkpoint_every: Optional[int] = None,
     # misc args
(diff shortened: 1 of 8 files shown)


projects/sandbox/train/train/validation.py
```diff
@@ -245,299 +50,220 @@ class Recorder:
...
-        glitch_preds = self.get_predictions(self.glitch_loader, model)
-        signal_preds = self.get_predictions(self.signal_loader, model)
-
-        return self.recorder(
-            model, train_loss, background_preds, glitch_preds, signal_preds
-        )
+        model.eval()
+        self._injection_idx = 0
+
+        # do the 0-shift inference manually since
```
EthanMarx, line 228: Probably should just skip the 0-shift right. We are collecting background so no need in obfuscating that. Also, in general, may be some signals hidden there.

projects/sandbox/train/train/validation.py
```diff
@@ -245,299 +50,220 @@ class Recorder:
...
-    background[num_h1:-num_coinc, 1] = l1_glitches[:, slc]
-    background[-num_coinc:] = coinc[:, :, slc]
+    def __post_init__(self):
+        self.auroc = BinaryAUROC(max_fpr=self.max_fpr)
+        self.duration = self.background.shape[-1] / self.sample_rate
+        self.kernel_size = int(self.kernel_length * self.sample_rate)
+        self.stride_size = int(self.stride * self.sample_rate)
+        self.pool_size = int(self.pool_length / self.stride)
+
+        integration_size = int(self.integration_length * self.sample_rate)
```
EthanMarx, line 139: This should be `int(self.integration_length * inference_sampling_rate)` (where `inference_sampling_rate = (1 / stride)` right?

EthanMarx:
Wan't to make sure I have this right based on the parameters you've chosen:

`valid_stride = 0.5` -->  `inference_sampling_rate` at 2 Hz.
`injection_stride = 4` --> Signals are injected every 4 seconds. So, for each signal we get 8 predictions, but, given our `inference_sampling_rate`, at most 2 of these predictions contain the coalescence time.
`integration_length = 1` --> With `inference_sampling_rate` of 2 Hz we are basically averaging every 2 predictions, which is the number we expect to get with the coalescence time in the window for each signal. 
`pool_length = 4` --> max pooling every 4 seconds. Presumably (or ideally) the two samples where the coalescence time is in the window will be the values pooled to.

projects/sandbox/train/train/utils.py
```diff
@@ -53,110 +35,137 @@ def split(X: Tensor, frac: float, axis: int) -> Tuple[Tensor, Tensor]:
...
-):
-    # build a glitch sampler from a pre-saved bank of
-    # glitches which will randomly insert them into
-    # either or both interferometer channels
-    glitch_dict = {}
-    valid_glitches_list = []
+def get_background(background_dir: Path):
+    # load in the data from the first segment in the background directory
+    try:
+        background_dataset = next(background_dir.iterdir())
```
EthanMarx, line 41: Can we confirm that these come sorted?

projects/sandbox/train/train/validation.py
alecgunny: Oh wow yeah great catch

alecgunny:
@EthanMarx your understanding is correct except for the signal portion. I don't do something like our test-time injections where I inject into the timeseries then window that timeseries. I just take the batch of already-windowed background data and then downsample it by a factor of `injection_stride // stride` (aka `injection_stride * inference_sampling_rate`) so that signals are effectively `injection_stride` apart. So for the injection batches, there's _always_ a signal in the prediction and therefore no need to pool.

projects/sandbox/train/train/utils.py
alecgunny: I don't think they do but since we only download one training background segment for now it shouldn't be a problem. But this should definitely be made more general when we overhaul the data picture.

projects/sandbox/train/train/validation.py
alecgunny: Yeah true great call

projects/sandbox/train/train/validation.py
```diff
@@ -245,299 +50,220 @@ class Recorder:
...
+        preds = torch.nn.functional.max_pool1d(
+            preds, self.pool_size, stride=self.pool_size
+        )
+        return preds[0, 0]
 
+    def inject(self, X: torch.Tensor) -> torch.Tensor:
+        X = X[:: self._injection_step]
+        start = self._injection_idx
+        stop = start + len(X)
+        waveforms = self.waveforms[start:stop]
```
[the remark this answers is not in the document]
alecgunny: Oh man... yeah they definitely weren't, should be updated now

projects/sandbox/train/train/validation.py
```diff
@@ -245,299 +50,218 @@ class Recorder:
...
-    start = h1_glitches.shape[-1] // 2 - kernel_size // 2
-    slc = slice(start, start + kernel_size)
+    def __post_init__(self):
+        self.auroc = BinaryAUROC(max_fpr=self.max_fpr)
+        self.duration = self.background.shape[-1] / self.sample_rate
+        self.kernel_size = int(self.kernel_length * self.sample_rate)
+        self.stride_size = int(self.stride * self.sample_rate)
+        self.pool_size = int(self.pool_length / self.stride)
+
+        integration_size = int(self.integration_length * self.stride)
```
EthanMarx, line 139: Making sure this doesn't slip by: this should be `int(self.integration_length / stride)` right?

EthanMarx:
Okay got it I see the logic now. So we get one prediction per signal where the signal lies in the center of the kernel

projects/sandbox/train/train/validation.py
alecgunny: Yup, this busted the first run after this push 😑

projects/sandbox/train/train/train.py
```diff
@@ -228,92 +212,115 @@ def main(
...
+    augmentor = augmentor.to(device)
 
-    # create full training dataloader
+    # Create full training dataloader.
+    # Use waveform dataset to dictate what
+    # an "epoch" should be, adding a factor
+    # to account for our sky parameter sampling
+    # and to balance compute vs. validation resolution
+    waveforms_per_batch = batch_size * waveform_prob
+    batches_per_epoch = 4 * len(waveforms) / waveforms_per_batch
```
wbenoit26, line 323: This needs to be an `int`

projects/sandbox/train/train/utils.py
```diff
@@ -35,113 +35,122 @@ def split(X: Tensor, frac: float, axis: int) -> Tuple[Tensor, Tensor]:
...
-    full_duration = train_stop - train_start
-    if valid_frac is not None:
-        train_duration = (1 - valid_frac) * full_duration
-        valid_start = train_start + train_duration
-    else:
-        train_duration = full_duration
 
+def get_glitches(
+    glitch_dataset: Path, ifos: List[str], end_time: Optional[float] = None
+) -> GlitchSampler:
```
wbenoit26, line 59: This doesn't actually return a `GlitchSampler`

alecgunny: Good catch, fixed

projects/sandbox/train/train/train.py
alecgunny: Ah had this fixed in another branch, great catch

PR #352, opened 2023-06-02
Mdc data updates

alecgunny:
I _think_ this should be everything we need to train on something like the MDC data. It implements:
- Chunked dataloading across multiple files, which is ~10x faster than using a vanilla torch Dataset with multiprocessing
- Local whitening using the previous `kernel_length - (fduration + 1)` seconds worth of data to fit a PSD, which is then applied to the last `fduartion + 1` seconds of the kernel before slicing off `fduration / 2` seconds at the edges.
- Getting rid of glitch sampling for the time being since it complicates the picture a bit (mostly around getting the data to do a run)

The only major remaining question is how with respect to which PSD do we want to measure the SNR for rescaling? One option that would rearrange things conceptually, but would be clean-ish from an implementation standpoint, would be to implement the PSD calculator as a separate module that gets computed in the `Augmentor`, and which is then used to rescale each of the sampled waveforms. This batch of PSDs would then need to be returned by the data generator so that the `Preprocessor` could use it to apply whitening (or we could build this into the `Augmentor` as well). At export time, we would just instantiate a new PSD calculator and whitener and build these into the ensemble on the server side.


EthanMarx:
I think a separate PSD calculator module makes sense, since it can be used for both whitening and snr rescaling. Is that what you were suggesting above? Also, is this not just the  `SpectralDensity` transform in `ml4gw`?

alecgunny:
@EthanMarx ok cool, implemented things that way. It's more or less the same, but there's some extra stuff for interpolating to the data's kernel length and limiting its impulse response/highpass filtering.

projects/sandbox/train/train/validation.py
```diff
@@ -117,9 +118,13 @@ class Validator:
     tracker: LocalTracker
     background: torch.Tensor
     waveforms: torch.Tensor
+    asd_estimator: Callable
+    whitener: Callable
```
EthanMarx, line 122: Where exactly is the validation data whitened? It doesn't look like this is used anywhere.

alecgunny: Yeah good catch this was screwing things up, should be fixed now

projects/sandbox/export/export.py
```diff
@@ -103,22 +105,24 @@ def export(
...
     logging.info(f"Loading parameters from {weights}")
     state_dict = torch.load(weights, map_location="cpu")
     nn.load_state_dict(state_dict)
-    kernel_length = preprocessor.whitener.kernel_length.item()
-    logging.info(
-        f"Model will be exported with input kernel length of {kernel_length}s"
-    )
     nn.eval()
 
+    background_length = kernel_length - (fduration + 1)
```
EthanMarx, line 116: Going to suggest that we parameterize by 
`background_length` (amount of data used for calculating PSD), `inference_window_length` (amount of whitened data fed to the ResNet model) and `fduration`. 
So that `kernel_length = background_length + inference_window_length + fduration`.

EthanMarx:
Closing as updates here were merged with #361

PR #367, opened 2023-07-17
Chunked glitch loader

EthanMarx:
Adds out of memory glitch loader, based off ml4gw/ml4gw#59. 

The glitch generation project will now save individual glitch files for each ifo, for each segment. The glitch loader will load in chunks of glitches from these files, and yield batches of glitches of size `glitch_prob * batch_size` for each ifo from these chunks. 

TODO:

- [ ] Add some sort of caching so we don't have to rerun glitch project
- [x] Incorporate into training pipeline
- [ ] tests


EthanMarx:
Have a training run going with these additions and seems to be learning. Seeing throughput of ~1300 samples/s on dgx which I think is slower by about 600 samples/s. Worth finding out where that bottleneck comes from.

EthanMarx:
@alecgunny @wbenoit26 was hoping to get some feedback:

In the code I currently have implemented, the glitch project will query `training_psd_length + 2 * kernel_length + fduration` seconds of data for each glitch. (This extra kernel_length factor is to ensure we can sample the `t0` of the glitch at both the front and end of the kernel). One drawback is that if we wan't to change the `training_psd_length` we have to re run the glitch project. I think we'll eventually arrive at a `training_psd_length` that we stick with, but until we get there we may wan't to try out a few. More importantly, this requires us to save more data to disk and waste gpu memory carrying around `training_psd_length` of data for each glitch. 

An alternative approach I (unfortunately) just thought of, is to just whiten `t0 - kernel_length` to `t0 + kernel_length` for all of the glitches up front in the glitch generation project, and then insert them into the batches as we've done before. This would remove the extra `training_psd_length` of data we keep around on disk, and eventually have to get on to GPU. Im not sure if this reduction in data removes the necessity of the asynchronous `GlitchLoader`, since we are now analyzing much longer stretches of data and thus will have many more glitches. 

Anyways, curious to hear thoughts / if i overlooked anything

wbenoit26:
Would you mind rebasing this on top of `mdc-dev`? It's hard to tell what all the changes are. The latter idea you suggested makes sense to me conceptually, but what about the idea of weighting glitch indices in the background data more heavily? Is the concern that we'll have glitches that aren't in the background?

EthanMarx:
Yeah can rebase. Now that we have the chunked data loader for background segments, we no longer need to generate glitches outside of the background segments. 

The problem with the index over sampling is mostly difficulty implementing it with the chunked data loader. It can be done, but would require significant reworking of the loader in ml4gw. I found this solution simpler.

PR #373, opened 2023-07-21
Validation fixes

alecgunny:
Adding multiple-view inference on signals and commenting our SNR thresholding right now during validation. Also fixes a couple bugs to get training running and reduces batch size to be able to fit the 1.5s kernel on a V100, and increases batches per epoch to balance out the longer validation time.

Need to do some actual testing to make sure my implementation is doing what I think it does. It seems to be correct, but the convergence behavior is very different. This is also very possibly because of the different kernel size/hyperparameters, so will be worth tracking these things down before launching a hyperparameter search run.

### DIFF
--- a/projects/sandbox/pyproject.toml
+++ b/projects/sandbox/pyproject.toml
@@ -139,7 +139,7 @@ highpass = "${base.minimum_frequency}"
 psd_length = "${base.training_psd_length}"
 
 # augmentation args
-trigger_distance = -0.5
+trigger_distance = -0.75
 waveform_prob = 0.5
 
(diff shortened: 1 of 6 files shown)


EthanMarx:
Is there a reason we need to comment out SNR thresholding? I would expect setting the threshold to 0 would just keep all the waveforms

alecgunny:
@EthanMarx yeah the issue is that we also use that same threshold as the min for our SNR sampling distribution. On the one hand coupling these _sort of_ makes sense but maybe it's more sensible to introduce a separate arg for this

PR #374, opened 2023-07-23
Chunked glitch loader

EthanMarx:
Adds a chunked glitch loader


projects/sandbox/train/train/train.py
```diff
@@ -322,6 +326,25 @@ def main(
...
+
+    # find glitch paths and initiate glitch loader
+    glitch_paths = {}
+    for ifo in ifos:
+        ifo_dir = glitch_dir / ifo
+        glitch_paths[ifo] = list(ifo_dir.iterdir())
+
+    glitch_loader = ChunkedGlitchDataset(
+        reads_per_chunk=10,
+        batches_per_chunk=int(batches_per_epoch / 4),
```
wbenoit26, line 339: Does the denominator need to match `chunks_per_epoch` below?

EthanMarx: Yeah I think `batches_per_chunk * chunks_per_epoch` (i.e. batches_per_epoch) should be the same for both the `background_loader` and `glitch_loader`

projects/sandbox/train/train/train.py
```diff
@@ -44,6 +46,8 @@ def main(
     highpass: float,
     # augmentation args
     waveform_prob: float,
+    glitch_prob: float,
```
wbenoit26, line 49: For consistency, I think we should rename this to `glitch_frac`, as we now have a guaranteed number of glitches per batch

wbenoit26:
No issues with the functionality that I can see. I am going to request that you update the docstrings in `background.py` and `glitches.py` to reflect any new/removed arguments and any changes to returned values.

wbenoit26:
Nice, looks good. What are your thoughts on tests? I know we need to redo the training tests, so do you want to add tests for this stuff when that happens?

EthanMarx:
Yeah I think adding tests to the train project will happen when we migrate the `ml4gw` `Whitener` and `ChunkedDataset`. So maybe this should wait til then, rebase, and add tests for all of these features together.

EthanMarx:
closing in favor of #462

PR #376, opened 2023-07-23
Added documentation to export

wbenoit26:


projects/sandbox/export/export/main.py
```diff
@@ -48,32 +53,46 @@ def main(
         architecture:
             A function which takes as input a number of witness
             channels and returns an instantiated torch `Module`
-            which represents a DeepClean network architecture
+            which represents an aframe network architecture
         repository_directory:
             Directory to which to save the models and their
             configs
-        outdir:
+        logdir:
```
alecgunny, line 60: Given that the logs are the only real output here, I support this name change. However, in that case, let's do away with the `weights is None` case inferring from this directory, since it's meant to be very different from the one where you would ever save model checkpoints. If `weights is None`, just save a randomly initialized model

projects/sandbox/export/export/main.py
```diff
@@ -48,32 +53,46 @@ def main(
...
             Path to save logs. If `weights` is `None`, this
             directory is assumed to contain a file `"weights.pt"`.
         num_ifos:
             The number of interferometers contained along the
             channel dimension used to train aframe
+        kernel_length:
+            Length of segment in seconds that the networks sees
```
alecgunny, line 67: "network sees"

projects/sandbox/export/export/main.py
```diff
@@ -174,7 +193,6 @@ def main(
             fduration=fduration,
             fftlength=fftlength,
             highpass=highpass,
-            name="snapshotter",
```
alecgunny, line 177: was there a reason to delete this? I guess this is probably the default name, but still worth making it clear here what's happening

alecgunny: Ah ok nvm seeing that it doesn't get used, nevermind then

alecgunny (commented):
Main blocker is the `weights is None` functionality. Curious to get your thoughts there

wbenoit26:
Yeah, I think that's sensible. Let me make that change and see how things look.

projects/sandbox/export/export/main.py
```diff
@@ -84,11 +83,10 @@ def main(
             during whitening
         highpass:
             Frequency to use for a highpass filter
-        weights:
+        weights_dir:
```
alecgunny, line 86: I think you should still be able to specify a full path to weights named whatever you like. Maybe do a check and if `weights.is_dir()` then `weights = weights / "weights.pt"`

libs/trainer/tests/test_trainer.py
```diff
@@ -3,7 +3,7 @@
 
 from aframe.trainer.trainer import train_for_one_epoch
 
-# test copied from deepclean
+# test copied from aframe
```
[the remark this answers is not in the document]
wbenoit26: True, should've actually looked at what I was replacing

projects/sandbox/export/export/main.py
wbenoit26: Agreed. I rearranged things a little, let me know what you think.

alecgunny (approved):

PR #378, opened 2023-07-23
Added documentation to trainer

wbenoit26:


libs/trainer/aframe/trainer/trainer.py
```diff
@@ -98,9 +100,12 @@ def train(
         train_dataset:
             An Iterable of (X, y) pairs where X is a batch of training
             data and y is the corresponding targets
-        valid_dataset:
-            An Iterable of (X, y) pairs where X is a batch of training
-            data and y is the corresponding targets
+        validator:
```
alecgunny, line 103: I wouldn't even necessarily dictate what this `Callable` needs to have. I would just something say "Callable object that takes as input a model and training loss value at each epoch and returns a flag indicating whether to terminate training or not."

alecgunny (commented):
Couple quick notes

alecgunny (approved):

PR #385, opened 2023-07-24
Added documentation to infer

wbenoit26:
@alecgunny When you get a chance, could you give me some text to describe `ChunkedSegmentLoader`? I don't think I understand it well enough to do a complete job of it.

### DIFF
--- a/projects/sandbox/infer/infer/callback.py
+++ b/projects/sandbox/infer/infer/callback.py
@@ -21,23 +21,20 @@ class Callback:
     Callable class for handling asynchronous server
     responses for streaming inference across sequences
     of timeseries data. New sequences should be
-    initialized by calling the `start_new_sequence`
-    method before starting to submit requests. Only
-    one sequence can be inferred upon at once.
+    initialized by calling the `initialize` method before
+    starting to submit requests. Only one sequence can be
+    inferred upon at once.
 
     Once inference has completed for each sequence,
     it will be asynchronously convolved with a
-    boxcar filter to perform local integration, then
-    both the raw and integrated timeseries will be
-    written to a file in `write_dir` with the filename
-    `out_{start}-{stop}.hdf5`, where `start` and `stop`
-    indicate the initial and final GPS timestamps of
-    the sequence.
+    boxcar filter to perform local integration, and then
+    the outputs will be clustered to find local maxima to
+    identify as events.
 
     Args:
-        write_dir:
-            Directory to which to save network outputs
-            in HDF5 format.
+        id:
+            Identifier used to match up sequences of
+            inference requests.
         inference_sampling_rate:
             Rate at which to sample inference windows
             from the input timeseries. Represents the
@@ -45,14 +42,24 @@ class Callback:
         batch_size:
             The number of subsequent windows to
             include in a single batch of inference.
-        window_length:
+        integration_window_length:
             Length of the window over which network
             outputs should be locally integrated,
+            specified in seconds.
+        cluster_window_length:
+            Length of the window over which network
+            outputs should be clustered, specified
+            in seconds
+        fduration:
+            Length of the time-domain whitening filter,
+            specified in seconds.
+        psd_length:
+            Length of background to use for PSD calculation,
             specified in seconds.
     """
 
     id: int
-    sample_rate: float
+    inference_sampling_rate: float
     batch_size: float
     integration_window_length: float
     cluster_window_length: float
@@ -84,7 +91,7 @@ class Callback:
             )
 
         duration = end - start
-        num_predictions = duration * self.sample_rate
+        num_predictions = duration * self.inference_sampling_rate
         num_steps = int(num_predictions // self.batch_size)
         num_predictions = int(num_steps * self.batch_size)
 
@@ -97,7 +104,7 @@ class Callback:
         self._started = {self.id: False, self.id + 1: False}
         # number of samples to remove from the beginning of the responses
         # due to the PSD burn-in
-        self.psd_size = int(self.sample_rate * self.psd_length)
+        self.psd_size = int(self.inference_sampling_rate * self.psd_length)
         return num_steps
 
     def integrate(self, y: np.ndarray) -> np.ndarray:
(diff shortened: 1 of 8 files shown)


projects/sandbox/infer/infer/callback.py
```diff
@@ -110,7 +117,9 @@ def integrate(self, y: np.ndarray) -> np.ndarray:
         integrated with 0s, so will have a lower magnitude
         than they technically should.
         """
-        window_size = int(self.integration_window_length * self.sample_rate)
+        window_size = int(
```
alecgunny, line 120: Oh crap was this what was being used? That's actually a potentially major bug

alecgunny: Ohhh nvm sorry saw that you renamed the variable, will comment there.

projects/sandbox/infer/infer/callback.py
```diff
@@ -21,38 +21,45 @@ class Callback:
...
+            Length of the time-domain whitening filter,
+            specified in seconds.
+        psd_length:
+            Length of background to use for PSD calculation,
+            specified in seconds.
     """
 
     id: int
-    sample_rate: float
+    inference_sampling_rate: float
```
alecgunny, line 62: On the one hand I agree that this helps distinguish what sampling rate this should refer to based on our nomenclature. The reason I opted to call this `sample_rate` in the first place is to emphasize that the output of the NN is just a timeseries with a different sample rate and physical significance. But I think I agree that belaboring that conceptual point is not worth the confusion in terms of code readability

projects/sandbox/infer/infer/data/batcher.py
```diff
@@ -1,11 +1,10 @@
-from typing import Iterator
-
 import numpy as np
+from infer.data import Injector
 from ratelimiter import RateLimiter
 
 
 def batch_chunks(
-    it: Iterator,
+    it: Injector,
```
alecgunny, line 7: Pedantic, but this doesn't need to be an `Injector` class: all it needs to be is an `Iterator` that returns two arrays (so `Iterator[np.ndarray, np.ndarray]` would probably be the most precise. How those arrays get generated is not this function's concern (though given the context it's maybe more helpful to specify precisely which iterator we typically expect)

alecgunny: Actually going through and reading the `Injector` code, this will actually never be an `Injector`: we `map` the `Injector` onto an `Iterable` to create an `Iterator` that returns the `Injector`'s action on each element of that `Iterable`. If someone were e.g. using `mypy`, it would raise a flag that `Injector` does not have a `__next__` method.

projects/sandbox/infer/infer/data/loader.py
```diff
@@ -43,6 +48,11 @@ def crawl_through_directory(
     sample_rate: float,
     shifts: Optional[List[float]],
 ):
+    """
+    Go through a directory, finding all files that match the given
+    pattern. Yield the data in those files in chunks, as well as the
```
alecgunny, line 53: Maybe specify what the pattern is (i.e. stem ending in <timestamp>-<duration>)

projects/sandbox/infer/infer/deploy.py
```diff
@@ -66,6 +79,10 @@ def get_num_shifts(data_dir: Path, Tb: float, shift: float) -> int:
 
 
 def get_ip_address() -> str:
+    """
+    Gets the ip address for the Triton server.
```
alecgunny, line 83: It's just trying to get the local, cluster-internal IP address. It's not necessarily related to Triton.

projects/sandbox/infer/infer/main.py
```diff
@@ -142,9 +200,13 @@ def main(
             timeseries on which to perform inference.
             Each HDF5 file in this directory will be used
             for inference.
-        write_dir:
+        output_dir:
```
alecgunny, line 203: We don't write NN outputs anymore. Should be where we write events.

alecgunny (commented):
For the `ChunkedSegmentLoader`, I would say something like:

```
"""
Iterable context which, upon entering, launches chunked loading
of the files in `data_dir` in an asychronous process that gets closed
upon context exit. The object returned by `__enter__` is a iterator that,
for each file in `data_dir`, returns a generator that iterates through that
file's content (possibly with relative shifts between channels) in chunks
of length `chunk_length`.

Example usage:
\```python
data_dir = Path("/path/to/data")
loader = ChunkedSegmentLoader(
    data_dir,
    channels=["H1", "L1"],
    chunk_length=1024,
    sample_rate=2048,
    shifts=[0, 1]
)
with loader:
    # async loading has launched
    for iterator in loader:
        # each file in data_dir has its own iterator
        for chunk in iterator:
            # note that this won't be true in general for the
            # last chunk in each file, which might be shorter
            # than the specified chunk length
            assert chunk.shape == (2, 2048 * 1024)
\```
"""
```

You can ignore the backslashes before the tick marks, I just needed them so GitHub wouldn't stop doing code rendering. Does that make more sense?

projects/sandbox/infer/infer/callback.py
wbenoit26: Gotcha, I see. Yeah, `sample_rate` comes a few times with slightly different contexts. There's the sample rate of the strain data vs the sample rate of the inference, and there's instances when it determines the sample rate of a time series vs instances where it's just descriptive. Worth thinking about when it comes to standardizing nomenclature.

projects/sandbox/infer/infer/data/batcher.py
wbenoit26: Ah, okay, I didn't understand how the mapping worked.

wbenoit26:
Yeah, that's awesome, thanks for writing it up

wbenoit26:
Actually, hold on, I'm gonna update the description of `shifts` based on what you said for `datagen`

wbenoit26:
Never mind, I think for the way `shifts` is used in this project, the description is fine as-is.

alecgunny (approved):

PR #387, opened 2023-07-26
Remove io library and updated envs

### LINKED ISSUE #366: Retire `libs/io` library
Now that we perform injections and inference on the fly, we no longer use or need the `io` library.

ricco-hub:
Closes #366


EthanMarx:
Would also be great to remove any references to `/libs/io` in the `filters.yaml` (for example, [here](https://github.com/ricco-hub/aframe/blob/remove-io-lib/.github/filters.yaml#L24C3-L24C17))

I also think it's been a while since we've updated the `filters.yaml` dependencies. For example I am seeing references to `/libs/base`, which is now deprecated. Would be nice to double check that these filters are up to date for each `lib` / `project` so that we are triggering tests appropriately.

EthanMarx (approved):

PR #389, opened 2023-07-28
Added documentation to train

wbenoit26:


wbenoit26:
In addition to adding documentation, I also generalized the validation time shifts to work for more than 2 interferometers. If we'd rather put that change in a different PR, I'm happy to split it out.

projects/sandbox/train/train/validation.py
```diff
@@ -153,31 +248,40 @@ def __post_init__(self):
         self._injection_step = int(self.injection_stride // self.stride)
 
     def steps_for_shift(self, shift: float):
+        """Compute the number of kernels that will be taken per shift"""
         shift = abs(shift)  # doesn't matter which direction
-        return (self.duration - shift - self.kernel_length) // self.stride + 1
+        max_shift = shift * (self.num_channels - 1)
```
alecgunny, line 253: Why is the max shift greater when there are more channels? Doesn't a given `shift` value represent the maximum shift, with one channel at that shift and all other channels at or below it?

alecgunny: ahhh ok because you don't want to shift any two IFOs by the same amount. So I guess the _most_ general way to handle it would be to enforce that the smallest shift you use is `num_channels -1`, and then for each shift you iterate through permutations such that
1. One IFO is always at 0
2. One IFO is always at the current shift
3. All other IFOs `i` are shifted by some amount `s_i`, `0 < s_i < shift`, such that `s_i != s_j` if `i != j`, for which I guess there would be `shift choose (num_ifos - 1)` combinations? Idk I hate combinatorics

Anyway, that's probably overkill for here. This should work fine for our purposes, we'll have plenty of time to solve this before we're likely to add an IFO anyway.

alecgunny (approved):

PR #391, opened 2023-07-31
Changes required to get pipeline running

### LINKED ISSUE #390: `export-model` generating incorrect NN
A recent run of `export-model` from `main` pointed at a trained network seems to produce a functionally random NN. Here's a comparison of inference on a background event performed using various combinations of a preprocessor and and NN hosted either locally with Torch or with Triton:

image

The lower sets of output timeseries are what I would expect (and since they're performed with Torch locally, I take them as ground truth), but the upper timeseries show that no matter how the data is preprocessed, the Triton-hosted NN produces nonsense outputs.

By contrast, here is an output timeseries produced by the same model hosted on Triton, but with export done from [this branch](https://github.com/alecgunny/bbhnet/tree/sandbox-environment)

image

Clearly, something funky is happening during NN export in `main`. I suspect this may be somehow related to changes made in #376 , but I will do a finer-grained commit search later today to confirm.

alecgunny:
Closes #390 . Still some minor issues keeping the pipeline from running end-to-end. Most notably
- parameter conversion during training waveform generation and removal of `cosmology` argument in config (specific to the MDC prior, but still worth having a functioning `main`)
- Gets rid of glitch generation step for now until it can be fixed and #374 gets merged
- Fixes the bug mentioned in #390 that functionally ignores the `norm_groups` parameter

Pipeline should be able to run end-to-end after this. Would be curious to have someone verify this

### DIFF
--- a/libs/architectures/aframe/architectures/resnet.py
+++ b/libs/architectures/aframe/architectures/resnet.py
@@ -79,7 +79,7 @@ def get_norm_layer(groups: Optional[int] = None) -> nn.Module:
             num_groups = None if groups is None else min(num_channels, groups)
             super().__init__(num_channels, num_groups)
 
-    return GroupNorm
+    return NormLayer
 
 
 def convN(
--- a/projects/sandbox/datagen/datagen/scripts/waveforms.py
+++ b/projects/sandbox/datagen/datagen/scripts/waveforms.py
@@ -8,6 +8,7 @@ from datagen.utils.injection import generate_gw
 from typeo import scriptify
 
 from aframe.logging import configure_logging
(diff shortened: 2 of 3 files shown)


wbenoit26 (approved):
Nice, looks good to me. I'll give it a run through once the `datagen` tests finish.

wbenoit26:
Currently failing due to lack of `chirp_distance` key in some of the priors used for testing. I don't mind merging this in anyway, given that we'll definitely be removing the conversion function from main, but up to you.

As an aside, this kind of thing is why I'm in favor of having the conversion to all parameters implemented in `main` as well as in `mdc` - having to navigate around which keys are in which priors feels overly restrictive.

PR #392, opened 2023-08-02
Fixed type annotations and a variable name

wbenoit26:
This fixes the only issues I ran into trying to run the pipeline end-to-end (not counting the known issues with the vizapp, and not counting the MDC project). Just a couple of bugs that got introduced during my documentation PRs.


wbenoit26:
Everything here was addressed in #393

PR #393, opened 2023-08-04
Return to endo3

wbenoit26:
Creating this PR for easy comparison with `main`. I think this covers everything that needs to be changed to revert back, but @alecgunny @EthanMarx I'd appreciate a once-over to make sure I haven't missed anything important.


projects/sandbox/export/export/snapshotter.py
```diff
@@ -57,6 +56,7 @@ def __init__(
         length += kernel_length + fduration
         self.size = int(length * sample_rate)
 
+        fftlength = kernel_length + fduration
```
EthanMarx, line 59: I like this idae. I think it might be better practice to keep `fftlength` as an argument to our script, annotate it as `None`, and then calculate the default if it is not passed.
```python
@scriptify
def main(
    ...
    fftlength: Optional[float] = None,
    ...
):
    fftlength = fftlength or kernel_length + fduration
```
This keeps the flexibility of passing it as an argument, and also lets the user know in the function signature that this is a parameter of the calculation. 

Up to you though, either way will work

EthanMarx:
I think this gets most of the job done. We'll also wan't to uncomment the snr thresholding that we do in validation. 

Might be worth doing a mini end to end data generation run (say, just do one short 10,000 second segment or so) to make sure things run, and to sanity check the distributions. Maybe @ricco-hub can put together his nice plots using this data?

projects/sandbox/export/export/snapshotter.py
wbenoit26: Yeah, I like that.

wbenoit26:
Yeah, I did a full data generation run last night. Located on LLO in `/home/william.benoit/aframe/endo3/datadir`

EthanMarx (approved):

ricco-hub:
Some distributions
image
image
image
image

PR #395, opened 2023-08-10
Fix background file used for PSD calc

EthanMarx:
Ensures that, given the same directory of background files, we pick the same file every time for calculating PSD used for hopeless snr rejection.

(this may or may not be the answer to discrepancies we've been seeing between consecutive runs, but regardless I would argue is desired behavior so we can have reproducible results)


EthanMarx:
@alecgunny @wbenoit26 quick PR I think is worth getting in now

wbenoit26 (approved):

PR #396, opened 2023-08-10
Fixed shift calculation

### LINKED ISSUE #394: Update `calc_shifts_required` to reflect not doing zero-lag shift
The calculation performed in [`calc_shifts_required`](https://github.com/ML4GW/aframe/blob/cfeff58377bd01d57ef6942aa4d89b5de5dfbddf/projects/sandbox/infer/infer/deploy.py#L42) assumes that we're doing inference over zero-lag data, which we're no longer doing. The documentation should be updated (the sum should go from i = 1 to N), and the plus sign in [this line](https://github.com/ML4GW/aframe/blob/cfeff58377bd01d57ef6942aa4d89b5de5dfbddf/projects/sandbox/infer/infer/deploy.py#L62) should be a minus.

wbenoit26:
Closes #394 

Fixes the bug in the shift calculation that sometimes led to an extra shift, and updates it to reflect not using the zero-lag data.

### DIFF
--- a/projects/sandbox/infer/infer/deploy.py
+++ b/projects/sandbox/infer/infer/deploy.py
@@ -42,11 +42,11 @@ def aggregate_results(output_directory: Path):
 def calc_shifts_required(Tb: float, T: float, delta: float) -> int:
     r"""
     Calculate the number of shifts required to generate Tb
-    seconds of background.
+    seconds of background, not using the zero-lag data.
 
     The algebra to get this is gross but straightforward.
     Just solving
-    $$\sum_{i=0}^{N-1}(T - i\delta) \geq T_b$$
+    $$\sum_{i=1}^{N}(T - i\delta) \geq T_b$$
     for the lowest value of N, where \delta is the
     shift increment.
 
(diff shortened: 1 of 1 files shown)


EthanMarx:
It looks like the correct calc was implemented in the `timeslide_waveforms` project however. I would expect the infer project to break since there wouldn't be any waveforms to inject. Is there an explanation?

wbenoit26:
...Huh. That's interesting. No, I don't have an explanation for that, but I can look into it.

wbenoit26:
Just confirming, for 5 shifts of a background of 1,600,000 seconds, we could have fit ~150,000 waveforms. Instead, we have ~120,000 waveforms, which makes sense for 4 shifts. So the correct number of waveforms were being generated, but somehow that doesn't break inference.

wbenoit26:
@EthanMarx I think it makes sense just to get this in, rather than generalize things so function isn't in two places. I'll make an issue for it though.

EthanMarx:
Yeah I agree, sounds good

EthanMarx (approved):

PR #400, opened 2023-08-15
computing vizapp weights up front

alecgunny:


projects/sandbox/vizapp/vizapp/pages/performance_summary/sensitive_volume.py
```diff
@@ -75,25 +84,37 @@ def initialize_sources(self):
...
+            self.weights[combo] = weight / norm
 
         self.line_source = ColumnDataSource(dict(x=self.x))
 
         band_x = np.concatenate([self.x, self.x[::-1]])
         self.band_source = ColumnDataSource(dict(x=band_x))
         self.update()
 
+    def get_err(self, V, dV):
+        denom = np.sqrt(4 * np.pi) * 3 * V
```
wbenoit26, line 115: This is correct as-is, but I think it's clearer if it's phrased as `denom = 4 * np.pi * distance ** 2` (like, making it explicitly 1 / (dV/dr)).

wbenoit26 (approved):

projects/sandbox/vizapp/vizapp/pages/performance_summary/sensitive_volume.py
alecgunny: Ahhh wow that's a much cleaner reformulation, great catch. Let me know how the rewrite looks and then I'll merge

PR #403, opened 2023-08-16
bilby edge case fix

wbenoit26:
As Alec (somehow) deduced, sampling just 1 item from a `PriorDict` with a conversion function results in any extra keys generated by the conversion function ending up in the returned sample.


projects/sandbox/datagen/datagen/scripts/timeslide_waveforms.py
```diff
@@ -154,6 +154,19 @@ def main(
...
     while n_samples > 0:
         params = prior.sample(n_samples)
+
+        # If a Bilby PriorDict has a conversion function, any
+        # extra keys generated by the conversion function will
+        # be added into the sampled parameters, but only if
+        # we're sampling exactly 1 time. This removes those
+        # extra keys
+        if n_samples == 1:
+            extra_keys = [
```
alecgunny, line 164: overkill, but might be simpler to just go `params = {k: params[k] for k in parameters}`

Might also be worth adding a TODO to ditch this if/when [this change](https://git.ligo.org/lscsoft/bilby/-/merge_requests/1286) makes it into the lib

wbenoit26: Yeah, no, good point, added both

wbenoit26: Forgot that `parameters` has some keys that params won't have, so I corrected for that. Also, I _gotta_ stop pushing changes to `datagen` without testing them just because they seem simple.

alecgunny (approved):

PR #411, opened 2023-08-25
Add `apply_vetos` funtion to `Ledger` `EventSet` and `TimeSlideEventSet` objects

EthanMarx:


libs/analysis/aframe/analysis/ledger/events.py
```diff
@@ -50,6 +50,21 @@ def significance(self, threshold: F, T: float) -> F:
...
+        # TimeSlideEventSet does not have shift information,
+        # so apply vetoes as if the shift is zero
+        mask = np.logical_and(
+            vetos[:, :1] < self.time, vetos[:, 1:] > self.time
+        )
+
+        # mark a background event as vetoed
+        # if it falls into _any_ of the segments
+        veto_mask = False
+        veto_mask |= mask.any(axis=0)
```
alecgunny, line 63: Am I missing something with why this would be any different than just setting `veto_mask = mask.any(axis=0)`?

EthanMarx: Nope. I think I copy pasted from somewhere without any thought.

alecgunny (approved):

PR #421, opened 2023-09-07
Updated inference sampling rate and waveform spacing

wbenoit26:
I believe these are the only lingering config updates we need to make, but it would be good if everyone gave the entire `pyproject.toml` a quick look to make sure there's nothing else we need to change.


alecgunny (approved):
Noted this on Slack, but will mention it here for posterity: my one note is that it could be worth lowering the hopeless SNR thresh given that we tend to underestimate SNRs relative to the local noise level at test time due to using the PSD from the whole segment. But otherwise this looks ready to ship to me.

PR #427, opened 2023-10-01
Static vizapp script for 1D and 2D histograms

ricco-hub:


wbenoit26:
There's a lot of commit history here that we're not going to want. It looks like all your most recent updates are in the last commit? If that's the case, I think you can do the following:

1. `checkout` back to `main`
2. Pull the most recent version of the upstream main (`git pull --recurse-submodules upstream main`)
3. Create a new branch just for these changes (`git checkout -b <new-branch-name>`)
4. Apply this commit to that new branch (`git cherry-pick 589b39f`). That number after `cherry-pick` is the commit hash of your most recent commit. I've never actually used `cherry-pick` before; I think it'll work, but let me know if it doesn't.
5. Push that new branch to your remote repo (`git push origin <new-branch-name>`) and make a new PR
6. Close this PR

Let me know if you have questions about any of this.

PR #429, opened 2023-10-03
added check for N=0

rafia17:


alecgunny:
These looks great. Couple points of feedback:
- It might be helpful to have some tests for these two. For the N=0 issue, you can use `unittest.mock.patch` to make `torch.rand` return all 0s to ensure the N=0 case gets triggered and then do some tests on that output to ensure it's what you expect. For the length issue, it should be enough to use a `pytest.raises` context and call the function with bad values to make sure the appropriate error is getting raised.
- I think the pre-commit checks are failing because you have unnecessary empty lines in the code. As a rule, we only allow double empty lines between top-level class and function definitions, not inside them.

alecgunny (changes requested):
See my notes in the comment

rafia17:
Thanks for the feedback:

- Regarding testing, I have tested both instances extensively in last few weeks. Printing out values for mask on every run and verifying that `if N < 0:` block only gets executed for the correct value of the mask and subsequently for N. 
-  Similarly tested for the psd_length such that the error gets raised only when psd_length is less than the window length, otherwise it doesn't get raised.
- Regarding pre-commit checks failing, I have checked yet again but cannot see any double empty lines in either augmentor.py or train.py. So I am very puzzled here.

PR #452, opened 2023-10-24
Cherry pick val improvments

EthanMarx:
Cherry pick commits that add multi segment validation


EthanMarx:
@wbenoit26 Quick PR that merges validation improvements into paper. I think once this is in should be good to re-launch SV over time. Since we can parallelize training over the GPUs, I'm going to add 2 more intervals for a total of 8.

wbenoit26 (approved):

PR #453, opened 2023-10-24
Seed longevity project and validation bug fixes

EthanMarx:
1. Adds seeding for the longevity project runs
2. Fixes minor bugs in new multi segment validation. This should make it's way into `main` too.


EthanMarx:
@wbenoit26 I think this is the last PR of this flurry over the past couple of days. Thanks for reviewing them so quickly. I've relaunched datagen with a seed, and will launch the training runs immediately after.

wbenoit26 (approved):

PR #454, opened 2023-10-29
Condorize SV calc for SV over time runs

EthanMarx:


EthanMarx:
@wbenoit26 This should be ready to go
Added two big things:

1. Condorized sv calc code for intervals to speed things up. Also now save the FARs corresponding to the given thresholds.
2. Added glitch generation for the intervals so we can investigate performance as function of glitch rate

projects/longevity/data_gen/environment.yaml
```diff
@@ -1,4 +1,4 @@
-name: aframe-longevity
+name: aframe-datagen
```
wbenoit26, line 1: Is this now identical to the environment in `sandbox`?

EthanMarx: Yeah was kind of a trick to not have to rebuild since now that we've added `omicron` the environments are identical

wbenoit26 (approved):

PR #458, opened 2023-11-04
add pre-commit hook to remove print statemetns

EthanMarx:
Adding in pre-commit hook that removes any print statements :)


wbenoit26 (approved):
Hah, I like it. I'm interested in taking a look at the [Ruff](https://github.com/astral-sh/ruff) repo this one links to, but not a high priority

PR #463, opened 2023-11-25
Online deployment

wbenoit26:
Re-did the online deployment code to work with the local whitening, and the event detection all seems to work.

- [x] Write strain ~(and whitened strain)~ from around events in addition to network output
- [ ] ~Center events in timeseries?~


wbenoit26:
For sure! We've had two triggers at the 1/day level, the first corresponding to [this](https://gracedb.ligo.org/superevents/S231126c/view/) superevent. The latter, at 1385025527, doesn't correspond to anything in GraceDB

wbenoit26:
Additional to-do: Ran in to `RuntimeError: Missing FrEndOfFile structure`, which ended the run. Need to figure out why that happens, or just add something to ignore it when it does.

alecgunny:
@wbenoit26 did this happen during `read_channel`? If so, sounds like another edge case of weird things that can happen when you try to read a frame (hey there's another entendre) that hasn't fully written yet. Basically all the checks in there are meant to address that exact problem, so could be worth just adding another check for that error message and sleeping for a tick.

wbenoit26:
Yeah, I think that's probably the right call. I just want to check whether it's something that gets fixed after a second, or if the file will need to be skipped entirely.

EthanMarx:
@wbenoit26 Another todo thats not necessary for this PR would be to query iDQ log likelihood channel for events. The iDQ channel should be distributed in the same frames as the strain. Taking a maximum +- 1s from the event time or something like that should be good.

wbenoit26:
@EthanMarx I think it would be good to get this into the repo for the review process

EthanMarx (approved):

PR #466, opened 2023-11-27
Datagen bns changes

rafia17:
Following changes are implemented:

scripts/waveforms.py:

- a signal_type variable is added. Based on its value, we will call generate_gw() for BBH and generate_gw_bns for BNS.
- the call for generate_gw_bns is made through concurrent.futures.ProcessPoolExecutor to speed up waveform generation
- some logging info is added to prompt when waveform generation is finished etc

utils/injection.py:

- generate_gw_bns() is added. Its similar to generate_gw() except that it gets rid of the wraparound of the coalescence to the front of the signal.


projects/sandbox/datagen/datagen/scripts/waveforms.py
```diff
@@ -93,17 +95,32 @@ def main(
...
-        params,
-        minimum_frequency,
-        reference_frequency,
-        sample_rate,
-        waveform_duration,
-        waveform_approximant,
-        detector_frame_prior,
-    )
+    if signal_type == "bns":
+        with ProcessPoolExecutor(140) as exe:
```
EthanMarx, line 99: I like the idea of parallelizing this, but I don't think this is doing what you expect: This will submit a single job that will generate all the requested waveforms in one process.

projects/sandbox/datagen/datagen/utils/injection.py
```diff
@@ -91,6 +95,67 @@ def generate_gw(
...
+def generate_gw_bns(
+    sample_params: Dict[List, str],
+    minimum_frequency: float,
+    reference_frequency: float,
+    sample_rate: float,
+    waveform_duration: float,
+    waveform_approximant: str,
+    detector_frame_prior: bool = False,
+):
+    padding = 1
```
EthanMarx, line 107: What is the motivation for this? In general should avoid "magic numbers" - maybe make this a parameter with a default value

projects/sandbox/datagen/datagen/utils/injection.py
```diff
@@ -91,6 +95,67 @@ def generate_gw(
...
+    for i, p in enumerate(sample_params):
+        polarizations = waveform_generator.time_domain_strain(p)
+        polarization_names = sorted(polarizations.keys())
+        polarizations = np.stack(
+            [polarizations[p] for p in polarization_names]
+        )
+
+        # just shift the coalescence to the left by 200 datapoints
+        # to cancel wraparound in the beginning
+        dt = -200
```
EthanMarx, line 144: Again, another magic number: why is this 200? Is there a first principles motivation for this?

projects/sandbox/datagen/datagen/utils/injection.py
```diff
@@ -91,6 +95,67 @@ def generate_gw(
...
+        polarizations = np.roll(polarizations, dt, axis=-1)
+
+        # cut off the first sec of the waveform where the wraparound occurs
+        padding_datapoints = padding * sample_rate
+        signals[i] = polarizations[:, int(padding_datapoints) :]
+        if i == (n_samples / 4):
+            logging.info("Generated polarizations     : {}".format(i))
+        elif i == (n_samples / 2):
+            logging.info("Generated polarizations     : {}".format(i))
```
EthanMarx, line 153: I would argue this reduces readability more than it helps with logging. If you wan't to keep track maybe something cleaner would be 
```python
# every 10th waveform
if not i % 10:
    # note the logging.debug so that it's only called if verbose=True.
    logging.debug(f"{i + 1} polarizations generated") 
```

projects/sandbox/datagen/datagen/utils/injection.py
```diff
@@ -91,6 +95,67 @@ def generate_gw(
...
+
+    logging.info("Finished Generated polarizations")
```
EthanMarx, line 155: "Finished Generated polarizations" --> "Finished generating polarizations"

projects/sandbox/train/train/augmentor.py
```diff
@@ -211,28 +211,30 @@ def forward(self, X):
...
         N = mask.sum().item()
-        responses = self.sample_responses(N, X.shape[-1], psds[mask])
-        responses.to(X.device)
 
-        # perform swapping and muting augmentations
-        # on those responses, and then inject them
-        responses, swap_indices = self.swapper(responses)
-        responses, mute_indices = self.muter(responses)
-        X[mask] += responses
+        if N > 0:
```
EthanMarx, line 215: Remind me when this edge case would be reached? In what instance would we not wan't to inject waveforms?

projects/sandbox/datagen/datagen/scripts/waveforms.py
```diff
@@ -93,17 +95,32 @@ def main(
...
+                params,
+                minimum_frequency,
+                reference_frequency,
+                sample_rate,
+                waveform_duration,
+                waveform_approximant,
+                detector_frame_prior,
+            )
+            signals = future.result()
+    else:
```
EthanMarx, line 111: Since this is going into a dedicated `bns` branch, I think it might make sense to just assume we are generating bns waveforms, and not have this `if else`. We can move to generalizing everything down the line.

EthanMarx (changes requested):
Nice stuff, left some comments and questions!

projects/sandbox/datagen/datagen/scripts/waveforms.py
rafia17: So this was a long discussion with Alec on a thread  that incidentally did not include you. I am going to forward that thread to you and hopefully you can see the entire convo.
The crux was that, using concurrent.futures did reduce the waveform generation from days to under 46 min on the hanford box.

projects/sandbox/datagen/datagen/utils/injection.py
rafia17: So there is the issue where thebilby generated waveforms have coalescence at timestamp=0 sec and that's on purpose I think. We need to move the coalescence to the end of the waveform at time stamp = 16 sec (lets say). To do this I played with several waveforms and found that if we roll the waveforms by 200 datapoints to the left, then we do get most of the coalescence at the very end. There are some ring-down remnants in some cases and to deal with that, we chop off the first sec of the waveform. So the padding is set to 1 sec and the waveform is generated "longer "by an  amount equal to padding.  After rolling and chopping off the first sec, the resultant waveform is of the intended length and the coalescence is nicely at the end of it.
I will add some comments to the code to explain this properly

projects/sandbox/datagen/datagen/utils/injection.py
rafia17: Please see above

projects/sandbox/datagen/datagen/utils/injection.py
rafia17: agree, will change

projects/sandbox/datagen/datagen/scripts/waveforms.py
rafia17: If we don't want to create a separate repo for BNS, it will be most ideal and efficient, if we take care of this now rather than later. And write code keeping in mind that it works seamlessly with the main pipeline. Otherwise it will be a huge issue later to merge bns branch into main aframe branch.

projects/sandbox/datagen/datagen/scripts/waveforms.py
alecgunny: Ethan's right. This is only helpful if you submit a job for each choice of parameters, submitting one job that generates waveforms for _all_ the parameters won't multiprocess anything, it will just generate all the waveforms in serial in one process.

rafia17: Yes, so I get your point. Not sure why it reduced the waveform generation time so drastically. Or could it be that the hanford box just behaved rather nicely at that particular run. It's kind of puzzling. 
So I will remove the concurrent.futures for now. If we run into bottlenecks in generating BNS in future, we can revisit at that time.

